PR #9, opened 2018-03-03
NN-556 First attempt at creating an OFFENDER_KEY_WORKER table using

Mjwillis:
flyway, a Hibernate mapped OffenderKeyworker entity and some basic
integration tests for the mapping and repository.


Mjwillis:
Lgtm, will change it to use identifier for pk
 probably will remove audit stuff as it’s stupid!

PR #17, opened 2018-03-08
change rest call param to use enum toString rather than name

lbennett-moj:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/KeyworkerService.java
```diff
@@ -52,7 +52,7 @@ public KeyworkerService(OffenderKeyworkerRepository repository) {
 
         UriComponentsBuilder builder = UriComponentsBuilder.fromUriString("/key-worker/{agencyId}/allocations");
 
-        allocationFilter.getAllocationType().ifPresent(at -> builder.queryParam("allocationType", at.name()));
+        allocationFilter.getAllocationType().ifPresent(at -> builder.queryParam("allocationType", at.toString()));
```
Quozzbat: Good spot. Strictly speaking, you need to use getTypeCode() here but as toString() returns bare type code at the moment, it's not a big deal.

PR #18, opened 2018-03-12
NN-561 migrate key worker auto alloc

Quozzbat:
Interim commit. Skeleton Key worker table and model introduced. Key classes of auto-allocation service migrated.

### DIFF
--- a/src/main/java/uk/gov/justice/digital/hmpps/keyworker/model/KeyworkerStatus.java
+++ b/src/main/java/uk/gov/justice/digital/hmpps/keyworker/model/KeyworkerStatus.java
@@ -1,0 +1,41 @@
+package uk.gov.justice.digital.hmpps.keyworker.model;
+
+import java.util.HashMap;
+import java.util.Map;
+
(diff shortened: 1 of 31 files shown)


src/main/java/uk/gov/justice/digital/hmpps/keyworker/model/KeyworkerStatus.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.Map;
+
+public enum KeyworkerStatus {
+    ACTIVE("ACT"),
+    INACTIVE("INA"),
+    UNAVAILABLE_ANNUAL_LEAVE("UAL"),
+    UNAVAILABLE_LONG_TERM_ABSENCE("ULT"),
+    UNAVAILABLE_NO_PRISONER_CONTACT("UNP"),
+    UNAVAILABLE_SUSPENDED("USU");
+
```
lbennett-moj: Just wondering if the API should master the status descriptions also (instead of the UI),  in case we eventually switch to a reference data table.

PR #19, opened 2018-03-12
NN-481 keyworker search

lbennett-moj:
nb (wip) will come back to tests when resttemplate has been refactored by @Quozzbat .   

NN-555 covers this task - not sure i have the rights to link it to NN-481


src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/EliteStaffLocationRoleDto.java
```diff
@@ -0,0 +1,35 @@
+package uk.gov.justice.digital.hmpps.keyworker.dto;
```
Quozzbat: This is a duplicate of StaffLocationRoleDto - we don't need both.

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/KeyworkerService.java
```diff
@@ -129,7 +149,7 @@ public void allocate(KeyworkerAllocationDto keyworkerAllocation) {
     public void allocate(OffenderKeyworker allocation) {
         Validate.notNull(allocation);
 
-        // This service method creates a new allocation record, therefore it will apply certain defaults automatically.
+        // This service method creates a new allocation record, therefore it will convertToKeyworkerDto certain defaults automatically.
```
Quozzbat: ?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/KeyworkerService.java
```diff
@@ -153,4 +173,50 @@ public void allocate(OffenderKeyworker allocation) {
     public List<OffenderKeyworker> getAllocationsForKeyworker(Long staffId) {
         return repository.findByStaffId(staffId);
     }
+
+    public Page<KeyworkerDto> getActiveKeyworkers(String agencyId, Optional<String> nameFilter, PagingAndSortingDto pagingAndSorting) {
```
Quozzbat: This can return Key workers that are not active (if local Key worker record exists and status is not active). I have a similar issue with getting unallocated offenders - it seriously complicates pagination if you start removing non-compliant records before responding so I think the 'decoration' approach is correct but the method would be better named something like 'getKeyworkersWithStatus' (or just 'getKeyworkers'). I would call this from auto-allocate process and just filter out the non-active Key workers.

src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/EliteStaffLocationRoleDto.java
lbennett-moj: yes - we both added - i'll remove this

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/KeyworkerService.java
lbennett-moj: could have sworn the staff search was returning active only! Rename coming up

src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/KeyworkerServiceController.java
```diff
@@ -248,4 +248,60 @@ public ResponseEntity allocate(
...
+            notes = "Search for key workers using firstname or lastname",
+            nickname="keyworkersearch")
+
+    @ApiResponses(value = {
+            @ApiResponse(code = 200, message = "OK", response = KeyworkerDto.class, responseContainer = "List"),
+            @ApiResponse(code = 400, message = "Invalid request.", response = ErrorResponse.class, responseContainer = "List"),
+            @ApiResponse(code = 404, message = "Requested resource not found.", response = ErrorResponse.class, responseContainer = "List"),
+            @ApiResponse(code = 500, message = "Unrecoverable error occurred whilst processing request.", response = ErrorResponse.class, responseContainer = "List")  })
+
+    @GetMapping(path="/{agencyId}/members")
```
lbennett-moj: I'm not happy with this endpoint name but would like to review when all endpoints in place

PR #22, opened 2018-03-13
NN-587 Add real "create allocation" service; WIP, tests and validation TODO

steverendell:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/repository/OffenderKeyworkerRepository.java
```diff
@@ -1,14 +1,27 @@
...
+import java.time.LocalDateTime;
 import java.util.List;
 
 public interface OffenderKeyworkerRepository extends CrudRepository<OffenderKeyworker,Long> {
 
     List<OffenderKeyworker> findByOffenderNo(String offenderNo);
 
+    @Modifying
+    @Query("update OffenderKeyworker set" +
+            "  active = 'n'," +
```
Quozzbat: uppercase N

src/main/java/uk/gov/justice/digital/hmpps/keyworker/repository/OffenderKeyworkerRepository.java
```diff
@@ -1,14 +1,27 @@
...
+            "  expiryDateTime = :expiryDate," +
+            "  deallocationReason = :deallocationReason" +
+            " where offenderNo = :offenderNo and active = 'Y'")
+    int deactivate(@Param("offenderNo") String offenderNo, @Param("deallocationReason")DeallocationReason deallocationReason, @Param("expiryDate")LocalDateTime expiryDate);
```
Quozzbat: An alternative, and probably more 'JPA' approach, is to get the record, update the object accordingly and then save() it.

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/KeyworkerService.java
```diff
@@ -132,12 +136,59 @@ public String startAutoAllocation(String agencyId) {
...
+      AND OB.OFFENDER_BOOK_ID = :bookingId
+      AND OB.ACTIVE_FLAG = 'Y'
+  WHERE SLR.SAC_STAFF_ID = :staffId
+    AND SLR.POSITION = 'AO'
+    AND SLR.ROLE = 'KW'
+    AND TRUNC(SYSDATE) BETWEEN TRUNC(SLR.FROM_DATE) AND TRUNC(COALESCE(SLR.TO_DATE,SYSDATE))
+    }*/
+
+        // Remove current allocation if any
+        repository.deactivate(newAllocation.getOffenderNo(), newAllocation.getDeallocationReason(), LocalDateTime.now());
```
Quozzbat: I'd rather it wasn't the responsibility of the allocate method to deallocate a previous allocation. I think it should check if there is an existing active allocation for the offender and throw an exception if there is. Let application logic outside this deal with deallocation.

PR #39, opened 2018-03-21
NN-612 Support retreival of KW from Keyworker DB or go to Elite2api N…

Mjwillis:
…OMIS DB if not yet migrated


keyworker-service.iml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Mjwillis: Sorry, will remove

PR #81, opened 2018-04-20
NN-702 Ensure that the correct security headers are set

JamieDraperUK:
- Includes strict-transport-security


JamieDraperUK:
At the moment this changeset causes a 403

PR #101, opened 2018-05-16
NN-765 Remove INACTIVE from drop down list of available KW to allocation

Mjwillis:
Sort within each capacity band

### DIFF
--- a/src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/KeyworkerDto.java
+++ b/src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/KeyworkerDto.java
@@ -7,10 +7,12 @@ import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
+import org.apache.commons.lang3.StringUtils;
 import org.hibernate.validator.constraints.NotBlank;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/KeyworkerDto.java
```diff
@@ -60,4 +62,11 @@
...
     @ApiModelProperty(value = "Key worker is eligible for auto allocation.")
     private Boolean autoAllocationAllowed;
+
+    /**
+     * @return fullname (last name first name in lower case)
+     */
+    public String getFullName() {
+        return StringUtils.lowerCase(StringUtils.join(Arrays.asList(lastName, firstName), " "));
+    }
 }
```
lbennett-moj: is this returned as part of the json?  Should it be formatted properly or excluded?

Mjwillis: We could put an ignore on it.  Perhaps I should move it to a private method.  ....

Mjwillis: @lbennett-moj I've push a new change as a static method

PR #102, opened 2018-05-17
Only alloc for KW_ADMIN to new role

Mjwillis:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/UserRolesMigrationService.java
```diff
@@ -68,10 +70,11 @@ private void migrateRoles(String prisonId, Set<String> rolesToMatch, Set<String>
      * Extract the set of all usernames contained in the supplied data structure.
      *
      * @param usernamesByRole usernames grouped by role membership.
+     * @param filterRoles roles to include
      * @return all the supplied usernames as a set.
      */
-    private Set<String> allUsernames(List<Pair<String, Set<String>>> usernamesByRole) {
-        return usernamesByRole.stream().map(Pair::getRight).reduce(new HashSet<>(), SET_UNION);
+    private Set<String> allUsernames(List<Pair<String, Set<String>>> usernamesByRole, Set<String> filterRoles) {
```
bell-pa: This method doesn't return all usernames now. Its more like 'usernamesForUsersHavingOneOfTheseRoles(usernamesByRole, rolesToMatch)'. Or some more concise and descriptive name that I can't think of.

PR #143, opened 2018-10-17
NN-1240 Key-worker Compliance and Projected Sessions for Individual Staff Member KW

Mjwillis:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/KeyworkerStatsController.java
```diff
@@ -59,4 +61,34 @@ public KeyworkerStatsDto getStatsForStaff(
...
+            nickname="getKeyworkerStats")
+
+    @ApiResponses(value = {
+            @ApiResponse(code = 200, message = "OK", response = PrisonStatsDto.class),
+            @ApiResponse(code = 400, message = "Invalid request.", response = ErrorResponse.class ),
+            @ApiResponse(code = 404, message = "Requested resource not found.", response = ErrorResponse.class),
+            @ApiResponse(code = 500, message = "Unrecoverable error occurred whilst processing request.", response = ErrorResponse.class) })
+
+    @GetMapping(path = "/{prisonId}")
+    public PrisonStatsDto getKeyworkerStats(
```
JamieDraperUK: I wonder if a better name would be getStatsForPrison, or getPrisonStats?

Mjwillis: Agreed with change

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/KeyworkerStatsService.java
```diff
@@ -26,22 +49,81 @@ public KeyworkerStatsDto getStatsForStaff(Long staffId, String prisonId, LocalDa
...
+
+            int projectedKeyworkerSessions = getProjectedKeyworkerSessions(applicableAssignments, staffId, prisonId, fromDate, nextEndDate);
+            final BigDecimal complianceRate = getComplianceRate(sessionsDone, projectedKeyworkerSessions);
+
+            Integer entryCount = usageGroupedBySubType.get(KEYWORKER_ENTRY_SUB_TYPE);
+
+            return KeyworkerStatsDto.builder()
+                    .projectedKeyworkerSessions(projectedKeyworkerSessions)
+                    .complianceRate(complianceRate)
+                    .caseNoteEntryCount(entryCount != null ? entryCount : 0)
```
JamieDraperUK: This is done differently to sessionCount (one is inlined the other is not), was this left there by mistake?

JamieDraperUK: I see, maybe you could make entryCount work like sessionCount?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/KeyworkerStatsService.java
```diff
@@ -26,22 +49,81 @@ public KeyworkerStatsDto getStatsForStaff(Long staffId, String prisonId, LocalDa
...
+        Validate.notNull(toDate, "toDate");
+
+        return PrisonStatsDto.builder().build();
+    }
+
+    private BigDecimal getComplianceRate(int sessionCount, int projectedKeyworkerSessions) {
+        BigDecimal complianceRate = PERCENT_OPERAND;
+
+        if (projectedKeyworkerSessions > 0)  {
+            complianceRate = new BigDecimal(sessionCount).setScale(4, BigDecimal.ROUND_HALF_UP)
```
JamieDraperUK: This looks like a nice API for working calculations. Would using this approach benefit the logic in getProjectedKeyworkerSessions?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/KeyworkerStatsService.java
Mjwillis: Yes I'll tidy this up afterwards. there is alot more to do here for batch so I'll be refactoring this in due cause.

PR #155, opened 2018-12-04
NN-1469 - Filter out non active key worker staff

Mjwillis:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/KeyworkerStatsService.java
```diff
@@ -127,7 +132,15 @@ public PrisonKeyWorkerStatistic generatePrisonStats(String prisonId) {
                     .sortOrder(SortOrder.ASC)
                     .build();
             ResponseEntity<List<StaffLocationRoleDto>> activeKeyWorkers = nomisService.getActiveStaffKeyWorkersForPrison(prisonId, Optional.empty(), pagingAndSorting, true);
-            log.info("There are currently {} active key workers in {}", activeKeyWorkers.getBody().size(), prisonId);
+
```
steverendell: do we have 2 meanings for 'active' here

Mjwillis: Yes, active in NOMIS but not active in keyworker

PR #161, opened 2019-01-28
NN-1539 - Fix API to be nicer when no data available

Mjwillis:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/KeyworkerStatsController.java
```diff
@@ -88,7 +88,7 @@ public KeyworkerStatSummary getPrisonStats(
             @RequestParam(value = "fromDate", required = false)
             @DateTimeFormat(iso = DateTimeFormat.ISO.DATE)
                     LocalDate fromDate,
-            @ApiParam(value = "End Date of Stats, optional, will chosse yesterday if not provided (in YYYY-MM-DD format)")
+            @ApiParam(value = "End Date of Stats (inclusive), optional, will chose yesterday if not provided (in YYYY-MM-DD format)")
```
petergphillips: should be `choose` :trollface:

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/KeyworkerStatsService.java
```diff
@@ -368,33 +378,36 @@ private PrisonStatsDto getPrisonStatsDto(String prisonId, CalcDateRange range, P
...
-                .complianceTimeline(compliance)
-                .avgOverallCompliance(getAverageCompliance(compliance.values()))
                 .build();
+
+        if (dailyStats != null) {
+            var kwSummary = new TreeMap<>(dailyStats.stream().collect(
+                    Collectors.groupingBy(s -> s.getSnapshotDate().with(weekAdjuster),
+                            Collectors.summingLong(PrisonKeyWorkerStatistic::getNumberKeyWorkerSessions))
+            ));
+            prisonStats.setKeyworkerSessionsTimeline(kwSummary);
```
petergphillips: this looks odd - you are using the builder to create prisonStats and then calling the setters in the object.
Better to create var of the builder and then call the methods on that here

Mjwillis: @petergphillips - duh! - changed it now

PR #170, opened 2019-05-22
NN-1918 IEP Role Access

bell-pa:
New endpoint for assigning  roles to, and removing roles from users who match given criteria.
Criteria are membership of named caseloads and roles.


src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/RoleMangementController.java
```diff
@@ -0,0 +1,55 @@
...
+import uk.gov.justice.digital.hmpps.keyworker.dto.RoleAssignmentsSpecification;
+import uk.gov.justice.digital.hmpps.keyworker.rolemigration.RoleAssignmentsService;
+
+@Api(tags = {"caseloads-roles"})
+
+@RestController
+@RequestMapping(value = "caseloads-roles")
+@Slf4j
+public class RoleMangementController {
+
```
bell-pa: Spelling!

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
```diff
@@ -0,0 +1,77 @@
...
+        this.roleService = roleService;
+    }
+
+    /**
+     * Look up users by caseload (prison) and membership of specification.rolesToMatch.
+     * for each matched user:
+     * Assign to the user the roles identified in specification.rolesToAssign.
+     * Remove from the user (at the caseload) the roles identified by specification.rolesToRemove.
+     */
+    @PreAuthorize("hasAnyRole('MAINTAIN_ACCESS_ROLES', 'MAINTAIN_ACESS_ROLES_ADMIN')")
```
petergphillips: `MAINTAIN_ACESS_ROLES_ADMIN` should be `ACCESS`?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
```diff
@@ -0,0 +1,77 @@
...
+
+            val usernamesForCaseload = findUsernamesMatchingRolesAtCaseload(specification.getRolesToMatch(), caseload);
+            log.debug("Found {} users for the {} caseload: {}.", usernamesForCaseload.size(), caseload, usernamesForCaseload);
+            usernamesForCaseload.forEach(username -> {
+                try {
+                    assignRolesToUser(username, specification.getRolesToAssign());
+                    try {
+                        removeRolesFromUserAtCaseload(caseload, username, specification.getRolesToRemove());
+                    } catch (Exception e) {
+                        log.debug("Failure while removing roles {} from user {} at caseload {}. Some roles may not have been removed.", specification.getRolesToRemove(), username, caseload);
```
petergphillips: bit naff catching `Exception` and then not logging the exception in any way.  What could cause the exception to be thrown?  Why is this `debug` - isn't this an `error` in our code somehow?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/RoleManagementController.java
```diff
@@ -0,0 +1,55 @@
...
+            nickname = "assignRolesFormUrlEncoded",
+            consumes = MediaType.APPLICATION_FORM_URLENCODED_VALUE
+    )
+    @ApiResponses({
+            @ApiResponse(code = 204, message = "Successful"),
+            @ApiResponse(code = 400, message = "Client error")
+    })
+    @PostMapping(consumes = MediaType.APPLICATION_FORM_URLENCODED_VALUE)
+    public ResponseEntity assignRolesForm(@ApiParam(value = "An application/x-www-form-urlencoded form.  Keys 'caseloads' and 'rolesToMatch' are mandatory, 'rolesToAdd' and 'rolesToRemove' should be supplied as needed") @RequestParam MultiValueMap<String, String> form) {
+        roleAssignmentsService.updateRoleAssignments(RoleAssignmentsSpecification.fromForm(form));
```
petergphillips: could this call `assignRolesJson` instead, which removes code duplication

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
```diff
@@ -0,0 +1,77 @@
...
+    @PreAuthorize("hasAnyRole('MAINTAIN_ACCESS_ROLES', 'MAINTAIN_ACESS_ROLES_ADMIN')")
+    public void updateRoleAssignments(@Valid final RoleAssignmentsSpecification specification) {
+        log.debug("Updating role assignments: {}", specification);
+        for (val caseload : specification.getCaseloads()) {
+
+            val usernamesForCaseload = findUsernamesMatchingRolesAtCaseload(specification.getRolesToMatch(), caseload);
+            log.debug("Found {} users for the {} caseload: {}.", usernamesForCaseload.size(), caseload, usernamesForCaseload);
+            usernamesForCaseload.forEach(username -> {
+                try {
+                    assignRolesToUser(username, specification.getRolesToAssign());
```
petergphillips: what will happen if the user already has the role - i.e. the code is re-run?

src/test/groovy/uk/gov/justice/digital/hmpps/keyworker/integration/specs/RoleAssignmentSpecification.groovy
```diff
@@ -0,0 +1,59 @@
...
+import org.springframework.http.HttpHeaders
+import org.springframework.http.HttpMethod
+import org.springframework.http.HttpStatus
+import org.springframework.http.MediaType
+import uk.gov.justice.digital.hmpps.keyworker.dto.RoleAssignmentsSpecification
+
+class RoleAssignmentSpecification extends TestSpecification {
+
+    def "A user that has the MAINTAIN_ACCESS_ROLES_ADMIN role can make role assignment changes"() {
+        given: "That I have the MAINTAIN_ACESS_ROLES_ADMIN role"
```
petergphillips: should be `ACCESS`

src/test/groovy/uk/gov/justice/digital/hmpps/keyworker/integration/specs/RoleAssignmentSpecification.groovy
```diff
@@ -0,0 +1,59 @@
...
+
+        HttpHeaders headers = new HttpHeaders()
+        headers.setBearerAuth(adminToken)
```
petergphillips: the `adminToken` doesn't have the `MAINTAIN_ACCESS_ROLES_ADMIN` role though :confused:, it has the `ROLE_MAINTAIN_ACCESS_ROLES` role

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
Mjwillis: Don't think MAINTAIN_ACCESS_ROLES should be allowed

src/test/groovy/uk/gov/justice/digital/hmpps/keyworker/integration/specs/TestSpecification.groovy
```diff
@@ -27,8 +27,8 @@ import static org.springframework.boot.test.context.SpringBootTest.WebEnvironmen
...
+    protected String token = "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJpbnRlcm5hbFVzZXIiOnRydWUsInVzZXJfbmFtZSI6IklUQUdfVVNFUiIsInNjb3BlIjpbInJlYWQiLCJ3cml0ZSJdLCJleHAiOjE4NDc4ODY4MTQsImF1dGhvcml0aWVzIjpbIlJPTEVfS1dfQURNSU4iLCJST0xFX09NSUNfQURNSU4iXSwianRpIjoiMzA1MjcwZjItYzZiYS00ZWY5LWIxOGMtYTllZTE1Zjk0ZTE2IiwiY2xpZW50X2lkIjoiZWxpdGUyYXBpY2xpZW50In0.WbqnPlPqr08tC7PEilLOrkzfh0UziyuD9lFnTUVbCQIK8Gp0c4iP9aJ0NydiPhcRzoUfaA9eUoZyMBp1GwpBa5kn-YNax4mBfL0pXT-zl6c1Bx6VqmVCNmAT5BXl2tkfAi0DnbRHHAqwaAR4UlGe5JSMt6mloomGqNFHkAuJinndR1LC9zlm0473LnLgmDaCVEuJDlUbaL660o5nHSEimC2yNpm8VI7_i5nYSzECpKw_BYkBuiuIQJp_9-BLcf67Q3FkxM1NDqQIPwfKtTQ2ayJJEWdaI0Z8PrwxTlDFwm7nfSBSf-rXSXaHaL_9BBnsdboOx5lEpl5p7f1A2bOjag"
```
Mjwillis: what has changed in token?

src/test/groovy/uk/gov/justice/digital/hmpps/keyworker/integration/specs/RoleAssignmentSpecification.groovy
Mjwillis: again that role should not allow access - remove ROLE_MAINTAIN_ACCESS_ROLES only the admin one

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
Mjwillis: Its also worth considering the restartability of this API.  ie. if one nomis call fails, do we stop, retry, etc.

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
bell-pa: Changed to log.warn(message, exception).

The exceptions are likely to be due to remote calls failing, not errors in the code.

src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/RoleManagementController.java
bell-pa: Wouldn't that be a bit weird?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
bell-pa: The end-point works by caseload, then by matched user within caseload.
The end-point calls elite2api end-points to add and remove roles from users (at caseloads).
The remote calls are repeatable; you can add the same role as many times as you like. Removing a role from a user who does not have the role results in a 404.
The end-point swallows the 404 errors and continues. 
More serious errors stop processing for the current user at the current caseload. Work resumes with the next user/caseload combination.  Adds are done before removes - so all roles should be added to a user before any roles are removed from that user.
Therefore the end-point can be used repeatedly with the same payload, and eventually, (on first call if the remote calls are all OK) the state of the role assignments should settle to what is specified in the payload.
I've tried various combinations of add/remove in different sequences against T3.  It seems to work reasonably well.
The logs should give plenty of information about what happened.
I can demo on Tuesday if that helps...

src/test/groovy/uk/gov/justice/digital/hmpps/keyworker/integration/specs/TestSpecification.groovy
bell-pa: I changed the access from private to protected so I could access the token in my tests.
Now I've reverted the change and created new tokens to use with the integration tests.

bell-pa:
1) Adding app insight tooling could be covered by a separate story.  
2) The api could return information about what it did. - summary or detail.
3) Could be re-worked to use Spring Batch or similar if needed.

PR #171, opened 2019-06-05
NN-1951 Raising an event to show results from the bulk role assignmen…

SimonMitchellMOJ:
…t process

returning results summary from the endpoint

### DIFF
--- a/src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/RoleManagementController.java
+++ b/src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/RoleManagementController.java
@@ -24,7 +24,7 @@ public class RoleManagementController {
 
     private final RoleAssignmentsService roleAssignmentsService;
 
-    public RoleManagementController(RoleAssignmentsService roleAssignmentsService) {
+    public RoleManagementController(final RoleAssignmentsService roleAssignmentsService) {
         this.roleAssignmentsService = roleAssignmentsService;
     }
 
@@ -39,7 +39,7 @@ public class RoleManagementController {
             @ApiResponse(code = 400, message = "Client error")
     })
     @PostMapping(consumes = MediaType.APPLICATION_JSON_VALUE)
-    public ResponseEntity<List<RoleAssignmentStats>> assignRolesJson(@RequestBody RoleAssignmentsSpecification specification) {
+    public ResponseEntity<List<RoleAssignmentStats>> assignRolesJson(@RequestBody final RoleAssignmentsSpecification specification) {
         val result = roleAssignmentsService.updateRoleAssignments(specification);
         return ResponseEntity.ok(result);
     }
@@ -55,7 +55,7 @@ public class RoleManagementController {
             @ApiResponse(code = 400, message = "Client error")
     })
     @PostMapping(consumes = MediaType.APPLICATION_FORM_URLENCODED_VALUE)
-    public ResponseEntity assignRolesForm(@ApiParam(value = "An application/x-www-form-urlencoded form.  Keys 'caseloads' and 'rolesToMatch' are mandatory, 'rolesToAdd' and 'rolesToRemove' should be supplied as needed") @RequestParam MultiValueMap<String, String> form) {
+    public ResponseEntity assignRolesForm(@ApiParam(value = "An application/x-www-form-urlencoded form.  Keys 'caseloads' and 'rolesToMatch' are mandatory, 'rolesToAdd' and 'rolesToRemove' should be supplied as needed") @RequestParam final MultiValueMap<String, String> form) {
         roleAssignmentsService.updateRoleAssignments(RoleAssignmentsSpecification.fromForm(form));
         return ResponseEntity.noContent().build();
     }
--- a/src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/RoleManagementController.java
+++ b/src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/RoleManagementController.java
@@ -7,11 +7,11 @@ import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
 import org.springframework.util.MultiValueMap;
 import org.springframework.web.bind.annotation.*;
-import uk.gov.justice.digital.hmpps.keyworker.dto.RoleAssignmentsSpecification;
 import uk.gov.justice.digital.hmpps.keyworker.dto.RoleAssignmentStats;
+import uk.gov.justice.digital.hmpps.keyworker.dto.RoleAssignmentsSpecification;
 import uk.gov.justice.digital.hmpps.keyworker.rolemigration.RoleAssignmentsService;
 
-import java.util.Map;
+import java.util.List;
 
 @Api(tags = {"caseloads-roles"})
 
@@ -35,11 +35,11 @@ public class RoleManagementController {
             consumes = MediaType.APPLICATION_JSON_VALUE
     )
     @ApiResponses({
-            @ApiResponse(code = 200, message = "Ok", response = RoleAssignmentStats.class, responseContainer = "Map"),
+            @ApiResponse(code = 200, message = "Ok", response = RoleAssignmentStats.class, responseContainer = "List"),
             @ApiResponse(code = 400, message = "Client error")
     })
     @PostMapping(consumes = MediaType.APPLICATION_JSON_VALUE)
-    public ResponseEntity<Map<String, RoleAssignmentStats>> assignRolesJson(@RequestBody RoleAssignmentsSpecification specification) {
+    public ResponseEntity<List<RoleAssignmentStats>> assignRolesJson(@RequestBody RoleAssignmentsSpecification specification) {
         val result = roleAssignmentsService.updateRoleAssignments(specification);
         return ResponseEntity.ok(result);
     }
--- a/src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/RoleManagementController.java
+++ b/src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/RoleManagementController.java
@@ -2,17 +2,23 @@ package uk.gov.justice.digital.hmpps.keyworker.controllers;
 
 import io.swagger.annotations.*;
 import lombok.extern.slf4j.Slf4j;
+import lombok.val;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
 import org.springframework.util.MultiValueMap;
 import org.springframework.web.bind.annotation.*;
 import uk.gov.justice.digital.hmpps.keyworker.dto.RoleAssignmentsSpecification;
+import uk.gov.justice.digital.hmpps.keyworker.dto.RoleAssignmentStats;
 import uk.gov.justice.digital.hmpps.keyworker.rolemigration.RoleAssignmentsService;
+
+import java.util.Map;
 
 @Api(tags = {"caseloads-roles"})
 
 @RestController
-@RequestMapping(value = "caseloads-roles")
+@RequestMapping(
+        value = "caseloads-roles",
+        produces = MediaType.APPLICATION_JSON_VALUE)
 @Slf4j
 public class RoleManagementController {
 
@@ -25,16 +31,17 @@ public class RoleManagementController {
     @ApiOperation(
             value = "Assign roles according to the provided specification",
             nickname = "assignRolesJson",
+            notes = "This returns a map of Prison ID to results of the role assignments",
             consumes = MediaType.APPLICATION_JSON_VALUE
     )
     @ApiResponses({
-            @ApiResponse(code = 204, message = "Successful"),
+            @ApiResponse(code = 200, message = "Ok", response = RoleAssignmentStats.class, responseContainer = "Map"),
             @ApiResponse(code = 400, message = "Client error")
     })
     @PostMapping(consumes = MediaType.APPLICATION_JSON_VALUE)
-    public ResponseEntity assignRolesJson(@RequestBody RoleAssignmentsSpecification specification) {
-        roleAssignmentsService.updateRoleAssignments(specification);
-        return ResponseEntity.noContent().build();
+    public ResponseEntity<Map<String, RoleAssignmentStats>> assignRolesJson(@RequestBody RoleAssignmentsSpecification specification) {
+        val result = roleAssignmentsService.updateRoleAssignments(specification);
+        return ResponseEntity.ok(result);
     }
 
     @ApiOperation(
--- a/src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/RoleAssignmentStats.java
+++ b/src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/RoleAssignmentStats.java
@@ -4,15 +4,15 @@ import io.swagger.annotations.ApiModelProperty;
 import lombok.Builder;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
+import lombok.ToString;
 
 import java.util.Map;
 
 @Builder
 @Getter
 @EqualsAndHashCode
+@ToString
 public class RoleAssignmentStats {
-
-    public enum Status {SUCCESS, FAIL, IGNORE}
 
     @ApiModelProperty(required = true, value = "Caseload")
     private String caseload;
@@ -29,38 +29,24 @@ public class RoleAssignmentStats {
     @ApiModelProperty(required = true, value = "Number of role unassignments failed")
     private long numUnassignRoleFailed;
 
-    public void addAssignResults(Map<Status, Long> assignResults) {
-        assignResults.forEach((status, count) -> {
-            switch (status) {
-                case SUCCESS:
-                    numAssignRoleSucceeded+= count;
-                    break;
-                case FAIL:
-                    numAssignRoleFailed+= count;
-                    break;
-                default:
-                    throw new IllegalStateException("Unexpected value: " + status);
-            }
-        });
-
+    public void incrementAssignmentFailure() {
+        numAssignRoleFailed++;
     }
 
-    public void addUnassignResults(Map<Status, Long> unassigningResults) {
-        unassigningResults.forEach(((status, count) -> {
-            switch (status) {
-                case SUCCESS:
-                    numUnassignRoleSucceeded+= count;
-                    break;
-                case FAIL:
-                    numUnassignRoleFailed+= count;
-                    break;
-                case IGNORE:
-                    numUnassignRoleIgnored+= count;
-                    break;
-                default:
-                    throw new IllegalStateException("Unexpected value: " + status);
-            }
-        }));
+    public void incrementAssignmentSuccess() {
+        numAssignRoleSucceeded++;
+    }
+
+    public void incrementUnassignmentSuccess() {
+        numUnassignRoleSucceeded++;
+    }
+    public void incrementUnassignmentFailure() {
+        numUnassignRoleFailed++;
+    }
+
+
+    public void incrementUnassignmentIgnore() {
+        numUnassignRoleIgnored++;
     }
 
     public Map<String, String> toMap() {
--- a/src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/RoleAssignmentStats.java
+++ b/src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/RoleAssignmentStats.java
@@ -1,0 +1,55 @@
+package uk.gov.justice.digital.hmpps.keyworker.dto;
+
+import io.swagger.annotations.ApiModelProperty;
+import lombok.*;
+import uk.gov.justice.digital.hmpps.keyworker.rolemigration.RoleAssignmentsService;
+
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+@Getter
+@ToString
+@EqualsAndHashCode
+public class RoleAssignmentStats {
+
+    @ApiModelProperty(required = true, value = "Number of role assignments succeeded")
+    private long numAssignRoleSucceeded;
+    @ApiModelProperty(required = true, value = "Number of role assignments failed")
+    private long numAssignRoleFailed;
+    @ApiModelProperty(required = true, value = "Number of role un-assignments succeeded")
+    private long numUnAssignRoleSucceeded;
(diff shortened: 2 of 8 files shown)


src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
```diff
@@ -32,48 +38,96 @@ public RoleAssignmentsService(RoleService roleService) {
...
-                    } catch (Exception e) {
-                        val message = String.format("Failure while removing roles %1$s from user %2$s at caseload %3$s. Some roles may not have been removed. Continuing with next user.",specification.getRolesToRemove(), username, caseload);
-                        log.warn(message, e);
-                    }
-                } catch (Exception e) {
-                    val message = String.format("Failure while assigning roles %1$s to user %2$s. No roles will be removed from this user. Continuing with next user.", specification.getRolesToAssign(), username);
-                    log.warn(message, e);
+            val results = new RoleAssignmentStats();
+
+            for (val username : usernamesForCaseload) {
```
Mjwillis: use `usernamesForCaseload.forEach( username -> {
...
}`

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
```diff
@@ -32,48 +38,96 @@ public RoleAssignmentsService(RoleService roleService) {
...
+        } catch (Exception e) {
+            val message = String.format("Failure while assigning roles %1$s to user %2$s. No roles will be removed from this user. Continuing with next user.", roleCodeToAssign, username);
+            log.warn(message, e);
+            return FAIL;
+        }
+    }
+
+    private void removeRolesFromUserAtCaseload(final RoleAssignmentStats results, final String caseload, final String username, final List<String> rolesToRemove) {
+        boolean hasFail = false;
+        for (val roleCodeToRemove : rolesToRemove) {
```
Mjwillis: streams please

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
```diff
@@ -32,48 +38,96 @@ public RoleAssignmentsService(RoleService roleService) {
...
+            if (!hasFail) {
```
Mjwillis: flip

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
andrewrlee: Think we deleted the wrong comment.

This needs to be a foreach (unless we resort to an atomic boolean)... though thinking about it we could possibly look at using `reduce` instead.

The deleted comment was about flipping the boolean round we can have a look at tomorrow.

andrewrlee: maybe something like: 
```java
        return !rolesToAssign.stream().reduce(false, (hasFailed, role) -> {
            if (hasFailed) {
                results.addAssignResult(FAIL);
                return true;
            }
            val status = assignRole(username, role);
            results.addAssignResult(status);
            return status == FAIL;
        }, Boolean::logicalAnd);
```
(if we are happy with streaming and side effects)

src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/RoleAssignmentStats.java
```diff
@@ -0,0 +1,55 @@
...
+                numUnAssignRoleSucceeded++;
+                break;
+            case FAIL:
+                numUnAssignRoleFailed++;
+                break;
+            case IGNORE:
+                numUnAssignRoleIgnored++;
+                break;
+            default:
+                throw new IllegalStateException("Unexpected value: " + status);
```
petergphillips: what scenario would this happen?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/RoleAssignmentStats.java
```diff
@@ -0,0 +1,55 @@
...
+@Getter
+@ToString
+@EqualsAndHashCode
+public class RoleAssignmentStats {
+
+    @ApiModelProperty(required = true, value = "Number of role assignments succeeded")
+    private long numAssignRoleSucceeded;
+    @ApiModelProperty(required = true, value = "Number of role assignments failed")
+    private long numAssignRoleFailed;
+    @ApiModelProperty(required = true, value = "Number of role un-assignments succeeded")
```
petergphillips: don't need the `-` since unassignments is a word

src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/RoleAssignmentStats.java
```diff
@@ -0,0 +1,55 @@
...
+    private long numUnAssignRoleSucceeded;
```
petergphillips: don't need a capital `A` since Unassign is a word instead

src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/RoleAssignmentStats.java
```diff
@@ -0,0 +1,55 @@
...
+
+import io.swagger.annotations.ApiModelProperty;
+import lombok.*;
+import uk.gov.justice.digital.hmpps.keyworker.rolemigration.RoleAssignmentsService;
+
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+@Getter
+@ToString
```
petergphillips: do you need these?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/RoleAssignmentStats.java
```diff
@@ -0,0 +1,55 @@
+package uk.gov.justice.digital.hmpps.keyworker.dto;
+
+import io.swagger.annotations.ApiModelProperty;
+import lombok.*;
+import uk.gov.justice.digital.hmpps.keyworker.rolemigration.RoleAssignmentsService;
+
+@Builder
```
petergphillips: you're not using `Builder` are you?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
```diff
@@ -32,48 +38,96 @@ public RoleAssignmentsService(RoleService roleService) {
      * Remove from the user (at the caseload) the roles identified by specification.rolesToRemove.
      */
     @PreAuthorize("hasAnyRole('MAINTAIN_ACCESS_ROLES_ADMIN')")
-    public void updateRoleAssignments(@Valid final RoleAssignmentsSpecification specification) {
+    public Map<String, RoleAssignmentStats> updateRoleAssignments(@Valid final RoleAssignmentsSpecification specification) {
         log.info("Updating role assignments: {}", specification);
+        val caseloadResults = new TreeMap<String, RoleAssignmentStats>();
         for (val caseload : specification.getCaseloads()) {
```
petergphillips: would be better to do a `specification.getCaseloads().stream().map()` instead here rather than a `for` loop, even though its not great to `stream` with side effects...
To keep it simple could even add the caseload to `RoleAssignmentStats` and then you can return a `List` instead

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
```diff
@@ -32,48 +38,96 @@ public RoleAssignmentsService(RoleService roleService) {
      * Remove from the user (at the caseload) the roles identified by specification.rolesToRemove.
      */
     @PreAuthorize("hasAnyRole('MAINTAIN_ACCESS_ROLES_ADMIN')")
-    public void updateRoleAssignments(@Valid final RoleAssignmentsSpecification specification) {
+    public Map<String, RoleAssignmentStats> updateRoleAssignments(@Valid final RoleAssignmentsSpecification specification) {
         log.info("Updating role assignments: {}", specification);
+        val caseloadResults = new TreeMap<String, RoleAssignmentStats>();
```
petergphillips: why a `TreeMap` here - do you mean a `LinkedHashMap`?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
```diff
@@ -32,48 +38,96 @@ public RoleAssignmentsService(RoleService roleService) {
...
-                        val message = String.format("Failure while removing roles %1$s from user %2$s at caseload %3$s. Some roles may not have been removed. Continuing with next user.",specification.getRolesToRemove(), username, caseload);
-                        log.warn(message, e);
-                    }
-                } catch (Exception e) {
-                    val message = String.format("Failure while assigning roles %1$s to user %2$s. No roles will be removed from this user. Continuing with next user.", specification.getRolesToAssign(), username);
-                    log.warn(message, e);
+            val results = new RoleAssignmentStats();
+
+            for (val username : usernamesForCaseload) {
+                var success = assignRolesToUser(results, username, specification.getRolesToAssign());
```
petergphillips: lots of mixing of `var` and `val` here.  My preference is `final var` since that's the only one that actually is final and also omits the type...

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
```diff
@@ -32,48 +38,96 @@ public RoleAssignmentsService(RoleService roleService) {
...
+                if (success) {
+                    removeRolesFromUserAtCaseload(results, caseload, username, specification.getRolesToRemove());
                 }
-            });
+            }
+            var infoMap = Map.of(
```
petergphillips: if you included `caseload` in the `RoleAssignmentStats` like I mentioned earlier then this complexity could be moved to that class instead

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
```diff
@@ -32,48 +38,96 @@ public RoleAssignmentsService(RoleService roleService) {
...
+                    "prisonId", caseload,
```
petergphillips: Nothing to do with your changes, but we seem to have a problem with our domain model.  Is this `prisonId` or is it `caseload`?  I only realised that we were dealing with prisons when I saw this line!

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
```diff
@@ -1,28 +1,34 @@
...
+
+import static java.util.stream.Collectors.toCollection;
+import static uk.gov.justice.digital.hmpps.keyworker.rolemigration.RoleAssignmentsService.Status.*;
 
 @Service
 @Slf4j
 @Validated
 public class RoleAssignmentsService {
 
+    public enum Status {SUCCESS, FAIL, IGNORE}
```
petergphillips: this enum seems to be unnecessary.  It is contained within this class and therefore just adds complexity, especially when we then have to cope with the enum not being set.  Couldn't there just be individual incrementors on the stats class instead?

src/test/groovy/uk/gov/justice/digital/hmpps/keyworker/integration/specs/RoleAssignmentSpecification.groovy
```diff
@@ -25,7 +26,15 @@ class RoleAssignmentSpecification extends TestSpecification {
 
         then: "The request is accepted"
 
-        response.statusCode == HttpStatus.NO_CONTENT
+        response.statusCode == HttpStatus.OK
+        def roleAssignmentStats = jsonSlurper.parseText(response.body)
+        def moorlandResults = roleAssignmentStats['MDI']
+        moorlandResults != null
+        moorlandResults.numAssignRoleSucceeded == 0
```
petergphillips: would be nicer if there was a single comparison of the stats object rather than each of the fields individually

src/test/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsServiceTest.java
```diff
@@ -155,7 +173,9 @@ public void failureDuringAddRolesPreventsRemoveRoles() {
         verify(roleService).removeRole("U2", "MDI", "X");
 
         verify(roleService, atLeastOnce()).findUsersForPrisonHavingRole(any(), any());
-        verifyNoMoreInteractions(roleService);
+
+        verifyEventRaised("MDI", "2", expected);
+        verifyNoMoreInteractions(roleService,telemetryClient);
```
petergphillips: normally with unit testing the rule is to only verify one thing.  This method seems to be testing everything - could the telemetry testing be moved to a different method?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
```diff
@@ -32,48 +38,96 @@ public RoleAssignmentsService(RoleService roleService) {
...
+            }
+            var infoMap = Map.of(
+                    "prisonId", caseload,
+                    "numUsersMatched", String.valueOf(usernamesForCaseload.size()),
+                    "numAssignRoleSucceeded", String.valueOf(results.getNumAssignRoleSucceeded()),
+                    "numAssignRoleFailed", String.valueOf(results.getNumAssignRoleFailed()),
+                    "numUnAssignRoleSucceeded", String.valueOf(results.getNumUnAssignRoleSucceeded()),
+                    "numUnAssignRoleIgnored", String.valueOf(results.getNumUnAssignRoleIgnored()),
+                    "numUnAssignRoleFailed", String.valueOf(results.getNumUnAssignRoleFailed()));
+            telemetryClient.trackEvent("UpdateRollAssignment", infoMap,null);
```
petergphillips: worth ticking the box in intellij to automatically format and correct imports - some of the changes have incorrect formatting

src/test/groovy/uk/gov/justice/digital/hmpps/keyworker/integration/specs/RoleAssignmentSpecification.groovy
andrewrlee: now comparing returned map

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
andrewrlee: added the settings and formatted this file

src/test/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsServiceTest.java
SimonMitchellMOJ: separated out test to verify call to telemetryClient

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
SimonMitchellMOJ: we have moved the enum out of the class but are still using it

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
SimonMitchellMOJ: caseload included in RoleAssignmentStats

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
SimonMitchellMOJ: no longer applicable as using list

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
SimonMitchellMOJ: we have added caseload to  RoleAssignmentStats  and are now returning a list

src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/RoleAssignmentStats.java
SimonMitchellMOJ: using builder with in tests

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
SimonMitchellMOJ: altered method to use stream reduce as suggested

src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/RoleAssignmentStats.java
SimonMitchellMOJ: we use the getters but have removed other annotations

src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/RoleManagementController.java
```diff
@@ -25,16 +31,17 @@ public RoleManagementController(RoleAssignmentsService roleAssignmentsService) {
...
-            @ApiResponse(code = 204, message = "Successful"),
+            @ApiResponse(code = 200, message = "Ok", response = RoleAssignmentStats.class, responseContainer = "List"),
             @ApiResponse(code = 400, message = "Client error")
     })
     @PostMapping(consumes = MediaType.APPLICATION_JSON_VALUE)
-    public ResponseEntity assignRolesJson(@RequestBody RoleAssignmentsSpecification specification) {
-        roleAssignmentsService.updateRoleAssignments(specification);
-        return ResponseEntity.noContent().build();
+    public ResponseEntity<List<RoleAssignmentStats>> assignRolesJson(@RequestBody RoleAssignmentsSpecification specification) {
+        val result = roleAssignmentsService.updateRoleAssignments(specification);
```
JamieDraperUK: Could be final.

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
```diff
@@ -32,48 +37,90 @@ public RoleAssignmentsService(RoleService roleService) {
...
+
+        telemetryClient.trackEvent("UpdateRollAssignment", results.toMap(), null);
+        return results;
+    }
+
+    private Map<RoleAssignmentStats.Status, Long> assignRolesToUser(final String username, final List<String> rolesToAssign) {
+        return rolesToAssign.stream().reduce(new HashMap<RoleAssignmentStats.Status, Long>(), (results, role) -> {
+            if (!results.containsKey(FAIL)) {
+                val status = assignRole(username, role);
+                results.compute(status, (key, count) -> count == null ? 1 : count + 1);
```
petergphillips: still seems very complicated with the `status`es.  Why can't you pass the `results` down and just individually increment as necessary rather than having to keep a map of all the successes / failures, pass it all back up and then loop over them all etc.

src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/RoleAssignmentStats.java
andrewrlee: It wouldn't... hence illegalstate :-) 

It was generated by intelliJ when we added the default case, it's quite a common idiom to add a handler for the default case. 

Can remove if you feel strongly about it?

andrewrlee: no longer relevant

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
andrewrlee: stayed consistent with the existing code so used `val`s

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
andrewrlee: moved to use `caseload`s everywhere in this feature - granted the codebase is quite inconsistent

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
andrewrlee: Did it like this originally as:
 * Parameters should normally be treated as immutable
 * Not a big fan of modifying in-memory state of an object across multiple locations.
  
Switched it to the suggested approach..

src/main/java/uk/gov/justice/digital/hmpps/keyworker/rolemigration/RoleAssignmentsService.java
andrewrlee: enum now removed

src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/RoleManagementController.java
andrewrlee: added finals everywhere... something like checkstyle would help avoid having to make these kinds of comments.

petergphillips: :+1:

PR #184, opened 2019-08-05
DT-129: Ensure only go back to nomis for agencies not migrated

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/KeyworkerService.java
```diff
@@ -146,7 +146,15 @@ private static String getKeyWorkerFullName(final KeyworkerDto keyworkerDto) {
                         .staffId(staffDetail.getStaffId())
                         .email(staffDetail.getEmail())
                         .build());
+
             }
+            return Optional.empty();
+        }
+        final var detail = nomisService.getPrisonerDetail(offenderNo, false);
```
petergphillips: don't think should need `admin` here to get prisoner details.  If we do then we won't go to nomis anyway...

PR #186, opened 2019-10-11
DT-283: Add in java agent and remove w3c interceptor

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/config/RestTemplateConfiguration.java
```diff
@@ -57,31 +54,23 @@ public RestTemplate elite2ApiHealthRestTemplate(final RestTemplateBuilder restTe
...
-                .additionalInterceptors(getRequestInterceptors())
+                .additionalInterceptors(new JwtAuthInterceptor())
                 .build();
     }
 
     private RestTemplate getHealthRestTemplate(final RestTemplateBuilder restTemplateBuilder, final String uri) {
         return restTemplateBuilder
                 .rootUri(uri)
-                .additionalInterceptors(getRequestInterceptors())
+                .additionalInterceptors(new JwtAuthInterceptor())
```
Mjwillis: The request interceptor is replaced, what does this change mean?

petergphillips: the method takes either a list or varargs, only need the JWT interceptor here as the W3C tracing one happens automatically with the java agent

src/main/java/uk/gov/justice/digital/hmpps/keyworker/config/RestTemplateConfiguration.java
```diff
@@ -57,31 +54,23 @@ public RestTemplate elite2ApiHealthRestTemplate(final RestTemplateBuilder restTe
...
                 .setConnectTimeout(healthTimeout)
                 .setReadTimeout(healthTimeout)
                 .build();
     }
 
-    private List<ClientHttpRequestInterceptor> getRequestInterceptors() {
-        return List.of(
-                new W3cTracingInterceptor(),
```
petergphillips: don't need the w3c tracing interceptor here

src/main/java/uk/gov/justice/digital/hmpps/keyworker/config/RestTemplateConfiguration.java
Mjwillis: cool

PR #197, opened 2019-11-05
DT-348 Remove Hourly Deallocation process and improve movement checks

Mjwillis:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/batch/DeallocationRoute.java
```diff
@@ -15,30 +13,19 @@
...
-        this.service = service;
-    }
-
     @Override
     public void configure() {
 
         if (StringUtils.isNotBlank(cronExpression)) {
             from(QUARTZ_UPDATE_STATUS_URI + cronExpression)
-                    .to(DIRECT_DEALLOCATION);
+                    .log("*** DEPRECATED: removed de-allocation process");
```
Mjwillis: Just commenting out for now, will need to remove the DB line in a future commit

src/main/resources/application-batch.properties
```diff
@@ -1,11 +1,11 @@
 quartz.enabled=true
-deallocation.job.cron=0 0 * ? * *
+deallocation.job.cron=0 30 23 ? * *
```
Mjwillis: Just going to run once a day for now- but will do nothing

src/main/resources/application.yml
```diff
@@ -7,6 +7,11 @@ spring:
     hibernate:
       ddl-auto: none
 
+    properties:
```
Mjwillis: This stops an annoying error on start up

PR #200, opened 2019-11-07
Dt 348 Handle de-alloc when admitted from court or other

Mjwillis:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/ReconciliationService.java
```diff
@@ -160,16 +161,19 @@ public void checkMovementAndDeallocate(final OffenderEvent offenderEvent) {
...
+
         // check that FROM agency is from where the key-worker / prisoner relationship resides and the TO agency is not the same prison!
-        if (!offenderKeyWorker.getPrisonId().equals(movement.getToAgency()) && offenderKeyWorker.getPrisonId().equals(movement.getFromAgency())) {
+        if (!offenderKeyWorker.getPrisonId().equals(movement.getToAgency())) {
             // check if its a transfer then its to another prison
             if ("TRN".equals(movement.getMovementType())) {
                 if (nomisService.isPrison(movement.getToAgency())) {
                     offenderKeyWorker.deallocate(movement.getCreateDateTime(), DeallocationReason.TRANSFER);
                 }
-            } else {
```
brightonsbox: Are there only three movement types? This change would mean that anything other than TRN, REL, ADM would no longer be treated as a release

PR #211, opened 2020-01-14
DT-447: Bump libraries and use new header

petergphillips:


build.gradle
```diff
@@ -17,6 +17,18 @@ repositories {
     mavenCentral()
 }
 
+def isNonStable = { String version ->
```
petergphillips: needed so that `./gradlew dependencyUpdates` ignores beta versions etc

PR #213, opened 2020-01-14
DT-447: Handle deletion event too

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/events/EventListener.java
```diff
@@ -1,53 +1,109 @@
...
-                event = objectMapper.readValue(message.get("Message"), OffenderEvent.class);
-            }
-        } catch (IOException e) {
-            log.error("Failed to Parse Message {}", requestJson);
+            final var message = objectMapper.readValue(requestJson, new TypeReference<Message>() {
+            });
+            return Optional.ofNullable(message).orElseThrow(() -> new RuntimeException(String.format("Missing message from %s", requestJson)));
+        } catch (final JsonProcessingException e) {
+            log.error("Failed to parse Message {}", requestJson, e);
+            throw e;
```
petergphillips: we now throw the exception when we get an error.  Previously we would swallow with a message.  This will mean that the message will get retried and put on the dead letter queue so that when we start monitoring the dead letter queue we can recognise the failures easier.

src/main/java/uk/gov/justice/digital/hmpps/keyworker/events/EventListener.java
```diff
@@ -1,53 +1,109 @@
...
+            return Optional.ofNullable(event).orElseThrow(() -> new RuntimeException(String.format("Missing offender event from %s", requestJson)));
+        } catch (final JsonProcessingException e) {
+            log.error("Failed to parse event {}", requestJson, e);
+            throw e;
+        }
+    }
+
+    @Getter
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    private static class Message {
```
petergphillips: this is far too much code even with lombok.  The kotlin versions are single line each...

src/main/java/uk/gov/justice/digital/hmpps/keyworker/events/EventListener.java
```diff
@@ -1,53 +1,109 @@
...
+            case "EXTERNAL_MOVEMENT_RECORD-INSERTED":
                 reconciliationService.checkMovementAndDeallocate(event);
-            } else if ("BOOKING_NUMBER-CHANGED".equals(event.getEventType())) {
+                break;
+            case "BOOKING_NUMBER-CHANGED":
                 reconciliationService.checkForMergeAndDeallocate(event);
-            }
+                break;
+            case "DATA_COMPLIANCE_DELETE-OFFENDER":
+                Preconditions.checkState(StringUtils.isNotBlank(event.getOffenderIdDisplay()), String.format("Found null offender id for %s", requestJson));
```
andrewrlee: meh.. don't need to use String format? takes [template](https://github.com/google/guava/blob/master/guava/src/com/google/common/base/Preconditions.java#L527) anyway?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/events/EventListener.java
andrewrlee: will this lead to the exception being logged multiple times? here and in some global error handling?

petergphillips: agreed, changed

src/main/java/uk/gov/justice/digital/hmpps/keyworker/events/EventListener.java
mikehalmamoj: Indeed.  How do you apply the Json annotations in Kotlin on one line?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/events/EventListener.java
petergphillips: agreed, changed

src/main/java/uk/gov/justice/digital/hmpps/keyworker/events/EventListener.java
andrewrlee: Java 14 adds a preview of records which have similar syntax to kotlin data classes: https://openjdk.java.net/jeps/359

mikehalmamoj: Operation Kill-Kotlin begins.

PR #214, opened 2020-01-14
DT-447 convert to kotlin

petergphillips:


src/test/java/uk/gov/justice/digital/hmpps/keyworker/services/ReconciliationServiceTest.java
```diff
@@ -167,7 +168,7 @@ void testCheckWhenTransferredOutDeallocationOccurs() {
         assertThat(offenderKw.getDeallocationReason()).isNull();
         assertThat(offenderKw.getExpiryDateTime()).isNull();
 
-        service.checkMovementAndDeallocate(OffenderEvent.builder().bookingId(-1L).movementSeq(1L).build());
+        service.checkMovementAndDeallocate(new OffenderEvent(-1L, 1L, null));
```
petergphillips: would normally call constructor fields by name when class converted to kotlin

src/main/java/uk/gov/justice/digital/hmpps/keyworker/events/EventListener.kt
```diff
@@ -0,0 +1,65 @@
...
+
+  @JsonIgnoreProperties(ignoreUnknown = true)
+  private data class MessageAttributes @JsonCreator constructor(@JsonProperty("eventType") val eventType: Attribute)
+
+  @JsonIgnoreProperties(ignoreUnknown = true)
+  private data class Attribute @JsonCreator constructor(@JsonProperty("Value") val value: String)
+
+  @JsonIgnoreProperties(ignoreUnknown = true)
+  data class OffenderEvent @JsonCreator constructor(
+      @JsonProperty("bookingId") val bookingId: Long?,
```
andymarke: Is it worth adding the annotation when the name matches the property?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/events/EventListener.kt
```diff
@@ -0,0 +1,65 @@
...
+      @JsonProperty("Message") val message: String,
+      @JsonProperty("MessageAttributes") val messageAttributes: MessageAttributes)
+
+  @JsonIgnoreProperties(ignoreUnknown = true)
+  private data class MessageAttributes @JsonCreator constructor(@JsonProperty("eventType") val eventType: Attribute)
+
+  @JsonIgnoreProperties(ignoreUnknown = true)
+  private data class Attribute @JsonCreator constructor(@JsonProperty("Value") val value: String)
+
+  @JsonIgnoreProperties(ignoreUnknown = true)
```
andymarke: Why do you need this ignore explicitly I'm sure I haven't added this in the probation update project?

andymarke: (I see you have taken the existing Java class that was perhaps over annotated)

src/main/java/uk/gov/justice/digital/hmpps/keyworker/events/EventListener.kt
petergphillips: agreed, will fix

src/main/java/uk/gov/justice/digital/hmpps/keyworker/events/EventListener.kt
petergphillips: yes, don't need it in probation project as using gson instead.  Will switch to gson

build.gradle
```diff
@@ -141,9 +153,11 @@ dependencies {
     implementation 'org.apache.commons:commons-lang3:3.9'
     implementation 'org.apache.commons:commons-text:1.8'
     implementation 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.10.2'
+    implementation 'com.fasterxml.jackson.module:jackson-module-kotlin:2.10.2'
```
mikehalmamoj: If we're switching to gson can we remove the jackson libs?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/events/EventListener.kt
```diff
@@ -0,0 +1,44 @@
...
+        require(!event.offenderIdDisplay.isNullOrBlank()) { "Found blank offender id for $requestJson" }
+        keyworkerService.deleteKeyworkersForOffender(event.offenderIdDisplay)
+      }
+    }
+  }
+
+  private data class Message(val Message: String, val MessageAttributes: MessageAttributes)
+  private data class MessageAttributes(val eventType: Attribute)
+  private data class Attribute(val Value: String)
+  data class OffenderEvent(val bookingId: Long?, val movementSeq: Long?, val offenderIdDisplay: String?)
```
mikehalmamoj: Losing those Jackson annotations is just so much nicer.  But are there are wider considerations moving to Gson?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/events/EventListener.kt
```diff
@@ -0,0 +1,44 @@
...
+  open fun eventListener(requestJson: String) {
+    val (message, messageAttributes) = gson.fromJson(requestJson, Message::class.java)
+    val eventType = messageAttributes.eventType.Value
+    log.info("Processing message of type {}", eventType)
+    val event = gson.fromJson(message, OffenderEvent::class.java)
+    when (eventType) {
+      "EXTERNAL_MOVEMENT_RECORD-INSERTED" -> reconciliationService.checkMovementAndDeallocate(event)
+      "BOOKING_NUMBER-CHANGED" -> reconciliationService.checkForMergeAndDeallocate(event)
+      "DATA_COMPLIANCE_DELETE-OFFENDER" -> {
+        require(!event.offenderIdDisplay.isNullOrBlank()) { "Found blank offender id for $requestJson" }
```
mikehalmamoj: I wonder if this class is the right place to be validating the message contents?  It's already breaking the OCP - needs to change every time we have a new event.

src/test/java/uk/gov/justice/digital/hmpps/keyworker/services/ReconciliationServiceTest.java
mikehalmamoj: Sounds like a good reason to convert OffenderEvent :smile:

petergphillips: `OffenderEvent` has been converted.  The problem is that java code only does method calls by position, not by name.  I would have to convert the `ReconciliationServiceTest` to make use of it though.

build.gradle
petergphillips: I'm not suggesting switching to gson for the spring boot api calls, just for the manual hydration in this class.  Our spring boot apps in all projects use jackson so would be worth keeping it that way.    The case notes to probation service follows the same pattern in using gson just for manual hydration.

src/main/java/uk/gov/justice/digital/hmpps/keyworker/events/EventListener.kt
petergphillips: no, since only affects this class

src/main/java/uk/gov/justice/digital/hmpps/keyworker/events/EventListener.kt
petergphillips: I wasn't a great fan of putting this here, but I copied the scenario from one of the other projects so left it as is.  Will move in next commit.

build.gradle
petergphillips: See https://www.callicoder.com/configuring-spring-boot-to-use-gson-instead-of-jackson/ for more information on integrating gson with spring boot

PR #240, opened 2020-03-27
DT-691 Apply the buffer max in memory size to all decoders

mikehalmamoj:
This was presenting in prod on Keyworker API endpoint `/key-worker/{prisonId}/offenders/unallocated`.  You can test the fix through Postman on T3.


mikehalmamoj:
NOTE: I've come to the conclusion that the buffer size problem might just be a symptom of the fix in PR #239 which is released in 2020-03-27.1797.9f25c9f.  I'm going to monitor that release first before applying this one.

PR #241, opened 2020-05-04
DT-801: Use published gradle plugin

petergphillips:


build.gradle
```diff
@@ -1,121 +1,19 @@
...
-    }
+    id "uk.gov.justice.hmpps.gradle-spring-boot" version "0.1.0"
 }
 
 configurations {
     compile.exclude  module: "tomcat-jdbc"
 }
 
-ext ['spring-security.version'] = '5.3.0.RELEASE' // Updated since spring-boot-starter-oauth2-resource-server-2.2.5.RELEASE only pulls in 5.2.2.RELEASE (still affected by CVE-2018-1258 though)
+ext ['spring-security.version'] = '5.3.1.RELEASE' // Updated since spring-boot-starter-oauth2-resource-server-2.2.5.RELEASE only pulls in 5.2.2.RELEASE (still affected by CVE-2018-1258 though)
```
mikehalmamoj, line 9: One for later, but as this is related to the Owasp dependency check mitigation I think this should have already been moved to the plugin...

build.gradle
```diff
@@ -1,121 +1,19 @@
...
 
 dependencies {
-    annotationProcessor "org.springframework.boot:spring-boot-configuration-processor"
+    annotationProcessor 'org.springframework.boot:spring-boot-configuration-processor'
     annotationProcessor 'org.projectlombok:lombok:1.18.12'
+
     testAnnotationProcessor 'org.projectlombok:lombok:1.18.12'
 
-    implementation 'org.springframework.boot:spring-boot-starter-web'
-    implementation 'org.springframework.boot:spring-boot-devtools'
```
mikehalmamoj, line 117: This isn't in the plugin... not required?

build.gradle
petergphillips: interesting, didn't realise that.  Will investigate in next PR :+1:

build.gradle
petergphillips: not sure if it was a general copy and paste from a different project anyway.  Didn't think that devtools did anything if it didn't have a ui?

PR #242, opened 2020-05-04
DT-801: Switch to kotlin gradle file

petergphillips:


build.gradle.kts
```diff
@@ -0,0 +1,79 @@
...
+  implementation("org.springframework:spring-webflux")
+  implementation("org.springframework.boot:spring-boot-starter-reactor-netty")
+
+  implementation("org.apache.camel:camel-spring-boot:2.25.1")
+  implementation("org.apache.camel:camel-quartz2:2.25.1")
+
+  implementation("org.springframework:spring-jms")
+  implementation("com.amazonaws:amazon-sqs-java-messaging-lib:1.0.8")
+
+  implementation("org.hibernate:hibernate-core")
```
petergphillips, line 34: spotted that we were using `java8` here by accident - been deprecated by `core` now for a few years - https://stackoverflow.com/questions/39761619/why-is-hibernate-java8-hibernate-5-x-jar-shown-deprecated

PR #256, opened 2020-06-30
DT-941 Don't log 404 exceptions just log the details

mikehalmamoj:
Exceptions appear in App Insights but not founds are expected, not exceptional

### DIFF
--- a/src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/ControllerAdvice.java
+++ b/src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/ControllerAdvice.java
@@ -15,6 +15,8 @@ import uk.gov.justice.digital.hmpps.keyworker.exception.PrisonNotSupportedExcept
 
 import javax.persistence.EntityNotFoundException;
 
+import static java.lang.String.format;
+
 
 @org.springframework.web.bind.annotation.RestControllerAdvice(
         basePackageClasses = KeyworkerServiceController.class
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/ControllerAdvice.java
```diff
@@ -24,7 +26,11 @@
 
     @ExceptionHandler(WebClientResponseException.class)
     public ResponseEntity<byte[]> handleException(final WebClientResponseException e) {
-        log.error("Unexpected exception", e);
+        if (e.getRawStatusCode() == 404) {
+            log.info(format("Not found for %s caused by %s", e.getClass().getName(), e.getMessage()));
```
petergphillips, line 30: since it is slf4j should be:
```
            log.info("Not found for {} caused by {}", e.getClass().getName(), e.getMessage());
```

petergphillips: Sorry, should have been clearer.  I was actually expecting a new 
```
    @ExceptionHandler(NotFound.class)
```
block rather than an if block on the parent class

petergphillips: maybe you could even add the class as an extra handler on the not found one i.e.
```
    @ExceptionHandler({EntityNotFoundException.class, NotFound.class})
```

mikehalmamoj: The not found exception is wrapped in a WebClientResponseException - that's what bubbles up to the ControllerAdvice.  So I don't think I can do this without rethrowing everywhere it appears?

mikehalmamoj: Or do you mean rethrow the not found from within ControllerAdvice... hmmm I think that would work.

petergphillips: I mean changing the `handleNotFoundException` method to:
```java

    @ExceptionHandler({EntityNotFoundException.class, NotFound.class})
    public ResponseEntity<ErrorResponse> handleNotFoundException(final Exception e) {
        log.info("Not found for {} caused by {}", e.getClass().getName(), e.getMessage());
        return ResponseEntity
                .status(HttpStatus.NOT_FOUND)
                .body(ErrorResponse
                        .builder()
                        .status(HttpStatus.NOT_FOUND.value())
                        .developerMessage(e.getMessage())
                        .build());
    }

```

PR #273, opened 2020-11-06
NN-3025:  convert to Kotlin

jamesak-oncloud9:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/Extensions.kt
```diff
@@ -0,0 +1,15 @@
+package uk.gov.justice.digital.hmpps.keyworker
+
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import kotlin.reflect.full.companionObject
+
+interface Logging
+
+inline fun <reified T : Logging> T.logger(): Logger = LoggerFactory.getLogger(getClassForLogging(T::class.java))
```
petergphillips, line 9: Not sure this adds anything to the code base.  You still need a `companion object` anyway and just obfuscates what's going on

src/main/java/uk/gov/justice/digital/hmpps/keyworker/batch/CamelQuartzConfigurationFactory.kt
```diff
@@ -0,0 +1,104 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import java.util.*
+
+/* requires postgres db - does not work with in memory database */
+@Configuration
+@ConditionalOnProperty(name = ["quartz.enabled"])
+internal class CamelQuartzConfigurationFactory(
+        @Autowired
```
petergphillips, line 18: kotlin formatting seems off here.  Should be using `ktlint` and getting it to format automatically.

src/main/java/uk/gov/justice/digital/hmpps/keyworker/config/JmsConfig.kt
```diff
@@ -94,4 +92,8 @@ open class JmsConfig {
     ))
     return awsSqsClient.getQueueUrl(queueName).queueUrl
   }
+
+  companion object: Logging {
+    val log = logger()
```
petergphillips, line 97: why have you made the `log` public?  I can't see this is any better than the code it replaced - it is just less characters rather than adding value

PR #274, opened 2020-11-09
NN-3025-1: convert to kotlin: KeyworkerServiceController

jamesak-oncloud9:


build.gradle.kts
```diff
@@ -1,80 +1,80 @@
 plugins {
-  id("uk.gov.justice.hmpps.gradle-spring-boot") version "1.0.6"
-  kotlin("plugin.spring") version "1.4.10"
-  kotlin("plugin.jpa") version "1.4.10"
+    id("uk.gov.justice.hmpps.gradle-spring-boot") version "1.0.6"
```
JamieDraperUK, line 2: Format still seems off. Did you apply those Kotlin settings to Intelij?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/KeyworkerServiceController.kt
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    private fun updateAndMigrate(prisonId: String?, migrate: Boolean? = false, autoAllocate: Boolean, capacity: Array<Int>?, kwSessionFrequencyInWeeks: Int? = 1): Prison {
+        if (capacity != null) {
+            Validate.isTrue(capacity.size == 2, "Two capacity values must be specified.")
+            prisonSupportedService.updateSupportedPrison(prisonId, autoAllocate, capacity[0], capacity[1], kwSessionFrequencyInWeeks)
+        } else {
+            prisonSupportedService.updateSupportedPrison(prisonId, autoAllocate)
+        }
+        if (migrate!!) {
```
JamieDraperUK, line 283: Might be able to get away with doing:

```
migrate?.let { 
      keyworkerMigrationService.migrateKeyworkerByPrison(prisonId)
      roleMigrationService.migrate(prisonId)
} 
```

jamesak-oncloud9: Do you mean?

```
     migrate?.let { migrateBool ->
            if(migrateBool) {
                keyworkerMigrationService.migrateKeyworkerByPrison(prisonId)
                roleMigrationService.migrate(prisonId)
            }
        }
```

src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/KeyworkerServiceController.kt
```diff
@@ -0,0 +1,291 @@
...
+    private val roleMigrationService: UserRolesMigrationService,
+    private val prisonSupportedService: PrisonSupportedService
+) {
+
+    companion object {
+        private val log = LoggerFactory.getLogger(this::class.java)
+    }
+
+    @ApiOperation(value = "Key workers available for allocation at specified prison.", notes = "Key workers available for allocation at specified prison.", nickname = "getAvailableKeyworkers")
+    @ApiResponses(value = [ApiResponse(code = 200, message = "OK", response = KeyworkerDto::class, responseContainer = "List"), ApiResponse(code = 400, message = "Invalid request.", response = ErrorResponse::class, responseContainer = "List"), ApiResponse(code = 404, message = "Requested resource not found.", response = ErrorResponse::class, responseContainer = "List"), ApiResponse(code = 500, message = "Unrecoverable error occurred whilst processing request.", response = ErrorResponse::class, responseContainer = "List")])
```
JamieDraperUK, line 67: Worth checking but I don't need the default `ApiResponse(code = 200, message = "OK", response = KeyworkerDto::class) ` response is needed as it's generated by default.

petergphillips: agreed, not required, not sure I'd fix as part of the conversion to kotlin though

src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/KeyworkerServiceController.kt
```diff
@@ -0,0 +1,291 @@
...
+
+    private fun updateAndMigrate(prisonId: String?, migrate: Boolean? = false, autoAllocate: Boolean, capacity: Array<Int>?, kwSessionFrequencyInWeeks: Int? = 1): Prison {
+        if (capacity != null) {
+            Validate.isTrue(capacity.size == 2, "Two capacity values must be specified.")
+            prisonSupportedService.updateSupportedPrison(prisonId, autoAllocate, capacity[0], capacity[1], kwSessionFrequencyInWeeks)
+        } else {
+            prisonSupportedService.updateSupportedPrison(prisonId, autoAllocate)
+        }
+        migrate?.let { migrateBool ->
+            if(migrateBool) {
```
petergphillips, line 284: this looks a lot messier than the original.  Why allow `migrate` to be null?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/KeyworkerServiceController.kt
```diff
@@ -0,0 +1,291 @@
...
+    fun addSupportedPrisonForAutoAllocation(
+        @ApiParam("prisonId") @PathVariable("prisonId") prisonId: String,
+        @ApiParam("migrate") @Param("migrate") migrate: Boolean?,
+        @ApiParam(name = "capacity", value = "standard and extended default keyworker capacities for this prison, comma separated, e.g. &capacity=6,9") @Param("capacity") capacity: Array<Int>?,
+        @ApiParam(name = "frequency", value = "default KW Session Frequency in weeks (default 1)") @Param("capacity") frequency: Int?
+    ): Prison {
+        return updateAndMigrate(prisonId, migrate, true, capacity, frequency)
+    }
+
+    private fun updateAndMigrate(prisonId: String?, migrate: Boolean? = false, autoAllocate: Boolean, capacity: Array<Int>?, kwSessionFrequencyInWeeks: Int? = 1): Prison {
```
petergphillips, line 276: have you got your nullability right here - why is `prisonId` allowed to be `null` if it is a path parameter?  Why supply a default to `migrate` when it is always passed as a parameter?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/KeyworkerServiceController.kt
```diff
@@ -0,0 +1,291 @@
...
+    @GetMapping(path = ["/{prisonId}/members"])
+    fun keyworkerSearch(
+        @ApiParam(value = "prisonId", required = true) @PathVariable("prisonId") prisonId: String,
+        @ApiParam(value = "Filter results by first name and/or last name of key worker. Supplied filter term is matched to start of key worker's first and last name.") @RequestParam(value = "nameFilter") nameFilter: Optional<String>?,
+        @ApiParam(value = "Filter results by status of key worker.") @RequestParam(value = "statusFilter") statusFilter: Optional<KeyworkerStatus>?,
+        @ApiParam(value = "Requested offset of first record in returned collection of allocation records.", defaultValue = "0") @RequestHeader(value = PagingAndSortingDto.HEADER_PAGE_OFFSET, defaultValue = "0") pageOffset: Long?,
```
petergphillips, line 212: you've set a default value here but still have the parameter as a `Long?`

src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/KeyworkerServiceController.kt
jamesak-oncloud9: because `migrate: Boolean?,` from the controller endpoint method is optional. Happy to go back to `if (migrate!!)` less noise in my view, and because it's defaulted in the method call it'll never throw an NPE

src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/KeyworkerServiceController.kt
jamesak-oncloud9: agree re `prisonId`
regarding `migrate` it's optional from the endpoint method but can default it in the controller method before passing it in

src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/KeyworkerServiceController.kt
jamesak-oncloud9: I'll fix for all the methods

.editorconfig
```diff
@@ -0,0 +1,3 @@
+
+[{*.kt,*.kts}]
+indent_size = 2
```
petergphillips, line 3: you're going to get knackered with this config since you're a mixed java / kotlin project and it will mess up your java imports etc.  Worth not checking it in and using the one from the new plugin instead.

petergphillips: although don't do it now otherwise it will create lots of other modifications, will have to be done in a separate PR to upgrade etc.

src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/KeyworkerServiceController.kt
jamesak-oncloud9: done

JamieDraperUK (approved):
LGTM

PR #280, opened 2020-11-12
NN-3025: default to false

jamesak-oncloud9:


jsrobertson (approved):
LGTM.  Thanks for fixing.

PR #282, opened 2020-11-25
DT-1347 provide mechanism to copy DLQ back to normal queue

Mjwillis:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/QueueAdminResource.kt
```diff
@@ -0,0 +1,45 @@
...
+
+@RestController
+@Validated
+@ConditionalOnExpression("{'aws', 'localstack'}.contains('\${sqs.provider}')")
+@RequestMapping("/queue-admin", produces = [MediaType.APPLICATION_JSON_VALUE])
+class QueueAdminResource(private val queueAdminService: QueueAdminService
+) {
+
+  @PutMapping("/purge-dlq")
+  @PreAuthorize("hasRole('QUEUE_ADMIN')")
```
Mjwillis, line 23: Added a new role type

mikehalmamoj (approved):

PR #288, opened 2021-01-12
NN-3110 deallocate offender on complex event

JamieDraperUK:


mikehalmamoj (approved):

src/main/java/uk/gov/justice/digital/hmpps/keyworker/events/EventListener.kt
```diff
@@ -32,6 +32,7 @@ open class EventListener(
       "EXTERNAL_MOVEMENT_RECORD-INSERTED" -> reconciliationService.checkMovementAndDeallocate(event)
       "BOOKING_NUMBER-CHANGED" -> reconciliationService.checkForMergeAndDeallocate(event.bookingId)
       "DATA_COMPLIANCE_DELETE-OFFENDER" -> keyworkerService.deleteKeyworkersForOffender(event.offenderIdDisplay)
+      "COMPLEX_OFFENDER" -> keyworkerService.deallocate(event.offenderIdDisplay)
```
Mjwillis, line 35: @JamieDraperUK are you sure this is what is going to be sent, it might just be a change of status not that is complex - need to speak to Richard Adams

PR #295, opened 2021-03-08
NN-3170 pave the way for the new complexity of need queue

JamieDraperUK:
We need to add a new queue that subscribes to domain events, however before that happens we need to make the offender events queue config and env variables more specific.

- Remove embedded localstack
- Make config and env more specific


src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/QueueAdminService.kt
```diff
@@ -12,12 +12,12 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression
 import org.springframework.stereotype.Service
 
 @Service
-@ConditionalOnExpression("{'aws', 'localstack'}.contains('\${sqs.provider}')")
+@ConditionalOnExpression("{'aws', 'localstack'}.contains('\${offender-events-sqs.provider}')")
 class QueueAdminService(
```
JamieDraperUK, line 16: This will need to change a bit more in future PRs as it's still generic. Maybe we could extend it to so that it takes a queue name as a path variable. Or we have two separate admin queue services.

src/test/java/uk/gov/justice/digital/hmpps/keyworker/events/EventListenerTest.java
```diff
@@ -76,13 +76,6 @@ void testDeleteEventEmpty() throws IOException {
         verifyNoInteractions(reconciliationService);
     }
 
-    @Test
```
JamieDraperUK, line 79: No longer needed as it'll come from the new queue.

src/main/java/uk/gov/justice/digital/hmpps/keyworker/events/EventListener.kt
```diff
@@ -32,7 +32,6 @@ open class EventListener(
       "EXTERNAL_MOVEMENT_RECORD-INSERTED" -> reconciliationService.checkMovementAndDeallocate(event)
       "BOOKING_NUMBER-CHANGED" -> reconciliationService.checkForMergeAndDeallocate(event.bookingId)
       "DATA_COMPLIANCE_DELETE-OFFENDER" -> keyworkerService.deleteKeyworkersForOffender(event.offenderIdDisplay)
-      "COMPLEX_OFFENDER" -> keyworkerService.deallocate(event.offenderIdDisplay)
```
JamieDraperUK, line 35: No longer needed as it'll come from the new queue.

src/main/java/uk/gov/justice/digital/hmpps/keyworker/config/JmsConfig.kt
```diff
@@ -23,9 +23,9 @@ import org.springframework.jms.support.destination.DynamicDestinationResolver
 import javax.jms.Session
 
 @Configuration
-@ConditionalOnExpression("{'aws', 'localstack'}.contains('\${sqs.provider}')")
+@ConditionalOnExpression("{'aws', 'localstack'}.contains('\${offender-events-sqs.provider}')")
 @EnableJms
-open class JmsConfig {
+class JmsConfig {
```
JamieDraperUK, line 28: Will need to rename this in future PR's

petergphillips (approved):
LGTM

PR #299, opened 2021-03-11
NN-3170 add queue admin

JamieDraperUK:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/OffenderEventsAdminService.kt
```diff
@@ -20,6 +20,15 @@ class OffenderEventsAdminService(
   @Value("\${offender-events-sqs.dlq.name}") private val dlqName: String
 ) : QueueAdminService(awsSqsClient, awsSqsDlqClient, queueName, dlqName)
 
+@Service
```
sp-watson, line 23: Not that important but the file name ought to be changed if it has all the queue's in

sp-watson (approved):

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/OffenderEventsAdminService.kt
JamieDraperUK: Will update in next PR

PR #302, opened 2021-03-12
NN-3227 handle duplicates gracefully

JamieDraperUK:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/KeyworkerAllocationProcessor.java
```diff
@@ -65,7 +65,25 @@ public KeyworkerAllocationProcessor(final OffenderKeyworkerRepository repository
     }
 
     public List<KeyworkerAllocationDetailsDto> decorateAllocated(final List<OffenderKeyworker> allocations, final List<OffenderLocationDto> allOffenders) {
-        final var allOffendersMap = allOffenders.stream().collect(Collectors.toMap(OffenderLocationDto::getOffenderNo, Function.identity()));
+
+        final var duplicates = allOffenders.stream()
```
sp-watson, line 69: We might be able to reduce the code using a Map mergeFunction

sp-watson (approved):

PR #310, opened 2021-03-23
NN-3207: Upgrade camel to 3.8

sp-watson:
The first task was to move the object logic from Camel to Kotlin.

After doing that the upgrade was easy.


src/main/java/uk/gov/justice/digital/hmpps/keyworker/config/RetryConfiguration.kt
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  @Bean
+  fun enableNomisRetryTemplate(): RetryTemplate {
+    val retryTemplate = RetryTemplate()
+    val retryPolicy = SimpleRetryPolicy()
+    retryPolicy.maxAttempts = 3
+    retryTemplate.setRetryPolicy(retryPolicy)
+    val backoffPolicy = ExponentialBackOffPolicy()
+    backoffPolicy.multiplier = 0.29
```
sp-watson, line 34: This doesn't make much sense to me, but I copied the way it worked before. I think we should just have 1 retry template.

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/KeyworkerStatsBatchService.kt
```diff
@@ -0,0 +1,48 @@
...
+      log.info("Stats completed for {}", prisonId)
+    } catch (e: Exception) {
+      log.error("Error occurred processing {}", prisonId)
+      keyworkerStatsService.raiseStatsProcessingError(prisonId, e)
+    }
+  }
+
+  private fun generatePrisonStatsWithRetry(prisonId: String): PrisonKeyWorkerStatistic {
+    return defaultRetryTemplate.execute(
+      RetryCallback<PrisonKeyWorkerStatistic, RuntimeException> {
```
sp-watson, line 43: I used the retryTemplate instead of annotation, as the annotation has to be applied to a public method. Using annotation would lead to 2 possible implementations, both having drawbacks:

 - Add the annotation to NomisService(Impl). This would make it far harder to test and force a retry for all callers, which may not be appropriate.
 - make a class just for this annotation.

src/test/java/uk/gov/justice/digital/hmpps/keyworker/services/KeyworkerStatsBatchServiceTest.java
```diff
@@ -0,0 +1,121 @@
+package uk.gov.justice.digital.hmpps.keyworker.services;
```
sp-watson, line 1: These are Java classes as the other tests are also Java, but maybe we should be using Kotlin for all new files?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/KeyworkerStatsBatchService.kt
```diff
@@ -0,0 +1,48 @@
...
+  private val prisonSupportedService: PrisonSupportedService,
+  private val defaultRetryTemplate: RetryTemplate
+) {
+  val log: Logger = LoggerFactory.getLogger(this.javaClass.name)
+
+  fun generatePrisonStats() {
+    try {
+      val prisonsWithId = prisonSupportedService.migratedPrisons
+      log.info("There are {} migrated prisons", prisonsWithId.size)
+      for (prison in prisonsWithId.stream()) {
```
JamieDraperUK, line 22: Don't change now but might be more consistent to use, if possible:`
    prisonsWithId.forEach {
        generatePrisonStatsForPrison(it.prisonId)
    }
}`

src/test/java/uk/gov/justice/digital/hmpps/keyworker/services/KeyworkerStatsBatchServiceTest.java
JamieDraperUK: If possible use Kotlin. You may find that you need to delombok some of the code.

src/test/java/uk/gov/justice/digital/hmpps/keyworker/services/NomisBatchServiceTest.java
```diff
@@ -0,0 +1,90 @@
...
+import uk.gov.justice.digital.hmpps.keyworker.dto.CaseloadUpdate;
+import uk.gov.justice.digital.hmpps.keyworker.dto.Prison;
+
+import java.util.List;
+import java.util.Map;
+
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.ArgumentMatchers.isA;
+import static org.mockito.ArgumentMatchers.isNull;
+import static org.mockito.Mockito.*;
```
JamieDraperUK, line 20: Wildcard import slipped in.

JamieDraperUK (approved):
LGTM, as a rule we don't add new Java code because eventually it'll all be converted.

src/test/java/uk/gov/justice/digital/hmpps/keyworker/services/NomisBatchServiceTest.java
sp-watson: Strange?!

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/KeyworkerStatsBatchService.kt
sp-watson: That's weird, because this is the one bit of code that IntelliJ corrected for me?!

PR #311, opened 2021-03-23
NN-3207: Camel update - improved formatting

sp-watson:
Converted new tests from Java to Kotlin.

Also used better forEach construct.


src/test/java/uk/gov/justice/digital/hmpps/keyworker/services/NomisBatchServiceTest.kt
```diff
@@ -0,0 +1,105 @@
...
+  @MockBean
+  private lateinit var nomisService: NomisService
+
+  @MockBean
+  private lateinit var telemetryClient: TelemetryClient
+
+  @Test
+  fun enableNomis_makesPrisonApiCalls() {
+    val prisons = listOf(MDI, LEI, LPI)
+    `when`(nomisService.allPrisons).thenReturn(prisons)
```
JamieDraperUK, line 38: Should be able to use whenever here. if not might be a good time to add the Kotlin mockito lib

JamieDraperUK (approved):
LGTM

PR #326, opened 2021-04-08
Nn 3300

JamieDraperUK:
- turn off complexity functionality in prod


src/main/java/uk/gov/justice/digital/hmpps/keyworker/events/ComplexityOfNeedEventProcessor.kt
```diff
@@ -21,14 +22,20 @@ data class ComplexityOfNeedChange(
 class ComplexityOfNeedEventProcessor(
   private val keyworkerService: KeyworkerService,
   private val telemetryClient: TelemetryClient,
-  @Qualifier("gson") private val gson: Gson
+  @Qualifier("gson") private val gson: Gson,
+  @Value("\${complexity_of_need_uri}") private val complexityOfNeedUri: String?
```
JamieDraperUK, line 26: This is only temporary.

sp-watson (approved):

PR #328, opened 2021-04-08
NN-3300:

JamieDraperUK:
- improve check


src/main/java/uk/gov/justice/digital/hmpps/keyworker/events/ComplexityOfNeedEventProcessor.kt
```diff
@@ -31,7 +32,7 @@ class ComplexityOfNeedEventProcessor(
   }
 
   fun onComplexityChange(message: String) {
-    if (complexityOfNeedUri == null) {
+    if (StringUtils.isNullOrEmpty(complexityOfNeedUri)) {
```
petergphillips, line 35: should use the kotlin version i.e. `complexityOfNeedUri.isNullOrEmpty()`

petergphillips (approved):
LGTM

PR #330, opened 2021-04-09
NN-3298:

JamieDraperUK:
- Place the access controls at the controller level so that the event processor can call deallocate


src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/KeyworkerServiceController.kt
```diff
@@ -456,6 +457,7 @@ class KeyworkerServiceController(
     ]
   )
   @PutMapping(path = ["/deallocate/{offenderNo}"])
+  @PreAuthorize("hasAnyRole('OMIC_ADMIN')")
```
JamieDraperUK, line 460: The deallocate method is only used in this controller and the complexity event processor.

sp-watson: Personally I think the inconsistent usage of this annotation between the controller and service is a maintenance issue, but it is already a problem in this service (and there is no obvious simple solution) so I will approve anyway.

sp-watson (approved):

PR #331, opened 2021-04-13
NN-3283: New endpoint to get allocation history summary.

sp-watson:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/KeyworkerServiceController.kt
```diff
@@ -437,6 +438,31 @@ class KeyworkerServiceController(
...
+        message = "Unrecoverable error occurred whilst processing request.",
+        response = ErrorResponse::class
+      )
+    ]
+  )
+  @PostMapping(path = ["/allocation-history/summary"])
+  fun getKeyWorkerHistorySummaryForPrisoners(
+    @RequestBody offenderNos: List<String>?
+  ): List<OffenderKeyWorkerHistorySummary> {
+    Validate.notEmpty<List<String>?>(offenderNos, "Please provide a list of Offender Nos.")
```
JamieDraperUK, line 462: Could you handle this using an annotation?

sp-watson: I don't know - I copied it from `getOffenderKeyworkerDetailsListPost` as it seemed like a good idea

JamieDraperUK (approved):
LGTM!

PR #335, opened 2021-04-21
NN-3323:

JamieDraperUK:
- Use client creds for complexity api

### DIFF
--- a/src/main/java/uk/gov/justice/digital/hmpps/keyworker/config/WebClientConfiguration.kt
+++ b/src/main/java/uk/gov/justice/digital/hmpps/keyworker/config/WebClientConfiguration.kt
@@ -44,7 +44,7 @@ class WebClientConfiguration(
   }
 
   @Bean
-  fun complexityOfNeedWebClient(authorizedClientManager: OAuth2AuthorizedClientManager?, builder: WebClient.Builder): WebClient{
+  fun complexityOfNeedWebClient(authorizedClientManager: OAuth2AuthorizedClientManager?, builder: WebClient.Builder): WebClient {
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/justice/digital/hmpps/keyworker/config/WebClientConfiguration.kt
```diff
@@ -44,25 +44,27 @@ class WebClientConfiguration(
   }
 
   @Bean
-  fun webClient(builder: WebClient.Builder): WebClient {
-    return builder
-      .baseUrl(elite2ApiRootUri)
-      .filter(addAuthHeaderFilterFunction())
+  fun complexityOfNeedWebClient(authorizedClientManager: OAuth2AuthorizedClientManager?, builder: WebClient.Builder): WebClient {
+    val oauth2Client = ServletOAuth2AuthorizedClientExchangeFilterFunction(authorizedClientManager)
+    oauth2Client.setDefaultClientRegistrationId("elite2-api")
```
sp-watson, line 49: i assume this is for convenience - cos it should ideally be keyworker-api as the client

sp-watson (approved):

src/main/java/uk/gov/justice/digital/hmpps/keyworker/config/WebClientConfiguration.kt
JamieDraperUK: Yeah, it's confusing. Once decoded the client id is is in fact keyworker-api. I just need to rename all the config so it aligns.

PR #336, opened 2021-04-21
DT-1949: ⬆️ Update to latest localstack

petergphillips:


docker-compose-all.yml
```diff
@@ -96,13 +96,12 @@ services:
...
   localstack:
-    # pinned to 10.4 due to https://github.com/localstack/localstack/issues/1992
-    image: localstack/localstack:0.10.4
+    image: localstack/localstack:0.12.9.1
     networks:
       - hmpps
     container_name: localstack
     ports:
-      - "4567-4597:4567-4597"
+      - "4566-4597:4566-4597"
```
sp-watson, line 104: Why the change of ports?

mikehalmamoj: Rather than a port for each service they have consolidated all services on the same port.

mikehalmamoj (approved):

PR #370, opened 2021-12-02
NN-3474: tidy up tasks from changing the stats to exclude high complexity

richardpopple:


helm_deploy/values-dev.yaml
```diff
@@ -22,4 +22,4 @@ env:
   JAVA_OPTS: "-Xmx512m"
   SPRING_SECURITY_OAUTH2_RESOURCESERVER_JWT_JWK_SET_URI: "https://sign-in-dev.hmpps.service.justice.gov.uk/auth/.well-known/jwks.json"
   COMPLEXITY_OF_NEED_URI: https://complexity-of-need-staging.hmpps.service.justice.gov.uk
-  PRISONS_WITH_OFFENDERS_THAT_HAVE_COMPLEX_NEEDS: "LEI,MDI"
+  PRISONS_WITH_OFFENDERS_THAT_HAVE_COMPLEX_NEEDS: "LEI,MDI,AGI,BZI,DHI,DWI,ESI,EWI,FHI,LNI,NHI,PFI,SDI,STI"
```
richardpopple, line 25: Request from Dom, to reflect the complex case loads in dev

helm_deploy/values-preprod.yaml
```diff
@@ -19,7 +19,7 @@ ingress:
 env:
   PRISON_URI_ROOT: "https://api-preprod.prison.service.justice.gov.uk"
   AUTH_URI_ROOT: "https://sign-in-preprod.hmpps.service.justice.gov.uk/auth"
-  JAVA_OPTS: "-Xmx1024m"
+  JAVA_OPTS: "-Xmx512m"
```
richardpopple, line 22: Did a heap dump on the servers and discovered that the heap wasn't that close to 512 - so put it back to 512 as it also looked like with all the other memory settings it might potentially get over 2Gb and be killed by kubernetes.

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/KeyworkerStatsService.java
```diff
@@ -477,28 +477,28 @@ private SummaryStatistic getSummaryStatistic(final PrisonKeyWorkerAggregatedStat
...
+                    .percentagePrisonersWithKeyworker(percentage(prisonStats.getNumPrisonersAssignedKeyWorker(), prisonStats.getTotalNumEligiblePrisoners()))
                     .numProjectedKeyworkerSessions((int) projectedSessions)
-                    .complianceRate(getComplianceRate(prisonStats.getNumberKeyWorkerSessions(), projectedSessions))
+                    .complianceRate(rate(prisonStats.getNumberKeyWorkerSessions(), projectedSessions))
                     .build();
         }
         return null;
     }
 
-    private BigDecimal percentagePrisonersWithKeyworker(final double numPrisonersAssignedKeyWorker, final double totalNumEligiblePrisoners) {
```
richardpopple, line 488: Made these methods static as they are really util methods that don't alter state, and it makes testing easier. Can change if people find it ugly.

richardpopple: I'd quite like to get rid of all the use of Double and move everything to BigDecimal - but that's going to be a bit of work - I've raised it in a tech dept ticket

sp-watson (approved):

PR #372, opened 2022-01-13
NN-3898: update api to use latest dps gradle spring boot version

richardpopple:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/config/SpringFoxConfiguration.kt
```diff
@@ -0,0 +1,97 @@
...
+  private val buildProperties: BuildProperties? = null
+
+  @Bean
+  fun api(): Docket? {
+    val apiInfo = ApiInfo(
+      "Key Worker API Documentation",
+      "API for accessing the Key Worker services.",
+      getVersion(), "", contactInfo(), "", "", emptyList()
+    )
+    val docket = Docket(DocumentationType.SWAGGER_2)
```
richardpopple, line 43: So I wanted to put this to OAS_30. However there are a couple of our documentation annotations which break this (and I couldn't find any easy way to fix it - and I suspect its breaking because its SpringFox. So I'm going to detail this in the ticket to upgrade to SpringDoc - because it took a while to track down which annotation was causing the problem (@apiResponse without a response class)

sp-watson (approved):

PR #395, opened 2022-04-28
NN-3914 springdoc v3

timoj58:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/BatchController.java
```diff
@@ -22,7 +24,7 @@
 import static uk.gov.justice.digital.hmpps.keyworker.batch.PrisonStatsRoute.DIRECT_PRISON_STATS;
 import static uk.gov.justice.digital.hmpps.keyworker.batch.UpdateStatusRoute.DIRECT_UPDATE_STATUS;
 
-@Api(tags = {"batch"}, hidden = true)
+@Tag(name = "batch")
```
sp-watson, line 27: Any reason for removing the "hidden" - I know this should never be called by users, just cron jobs

src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/AllocationHistoryDto.java
```diff
@@ -3,6 +3,7 @@
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 import io.swagger.annotations.ApiModelProperty;
```
sp-watson, line 5: Is this still used? This comment applies to most of these files

src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/BatchController.java
timoj58: Tag doesnt have a hidden option, all the end points are hidden though, they also have security settings that i assume do not work (you would need to log in to see them).  My assumption is a lof of the api references are cut and paste code, given the differences and un-used things

src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/AllocationHistoryDto.java
timoj58: no i will remove, thought i had deleted, perhaps not.  Was expecting the linter to remove, running reformat / lint again

src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/BatchController.java
sp-watson: yeah, ok

src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/AllocationHistoryDto.java
sp-watson: hmm.. you are right - the linter should have picked it up. This is one of the first APIs so i think it hasn't got the linter set up properly. That would be worth investigating some time.

sp-watson (approved):

PR #422, opened 2022-08-03
NN-4278 override default allowlist for dev due to some int tests

timoj58:


helm_deploy/values-dev.yaml
```diff
@@ -14,6 +14,8 @@ generic-service:
  ingress:
    host: keyworker-api-dev.prison.service.justice.gov.uk
 
+ allowlist: false
+
 ingress:
   enabled: true
```
andrewrlee, line 20: there should only be on ingress section

timoj58: Hi Andy, we are in the middle of migrating from v0.47 to v1.2 but due to this service not using generic service, which we want to upgrade, i have discovered we can do the upgrade and ingress migration using the old ingress, and the new v1.2 ingress with the weightings.  If you look in the template folder you will see there is a custom ingress (ie the old one pre generic service).

andrewrlee (approved):
LGTM

PR #427, opened 2022-10-27
NN-4591 add annotations

timoj58:


helm_deploy/keyworker-api/values.yaml
```diff
@@ -31,7 +31,8 @@ generic-service:
     tlsSecretName: keyworker-api-cert
     v1_2_enabled: true
     v0_47_enabled: false
-    nginx.ingress.kubernetes.io/configuration-snippet: |
+    annotations:
+     nginx.ingress.kubernetes.io/configuration-snippet: |
```
petergphillips, line 35: would be better to use 2 spaces instead of 1 for the snippet

timoj58: reformatted

petergphillips (approved):
LGTM

PR #452, opened 2023-05-15
NN-5081 bump up camel to 3.20.4

p-nik:


neilmendum:
Thanks for doing this! Please could you also remove the pinning of versions in the [Renovate config](https://github.com/ministryofjustice/keyworker-api/blob/75fd6aa5f149740268d824c212a2f73cbe2a70dd/renovate.json#L12-L31) so that it will now bump them for us going forward

neilmendum:
Addressed with https://github.com/ministryofjustice/keyworker-api/pull/455

PR #485, opened 2023-08-14
NN-5031 Removed exclude logback logs.

nagendramoj:


build.gradle.kts
```diff
@@ -4,8 +4,6 @@ plugins {
   kotlin("plugin.jpa") version "1.9.0"
 }
 configurations {
```
neilmendum, line 6: I think you can remove the whole `configurations` block

neilmendum (approved):
LGTM

PR #488, opened 2023-08-16
NN-5031 Starting and Complete logs added to the Cronjobs.

nagendramoj:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/BatchController.java
```diff
@@ -112,8 +115,9 @@ public List<Long> runBatchUpdateStatus() {
 
     @PostMapping(path = "/key-worker-recon")
     public void runKWReconciliation() {
-        log.info("Cronjob triggered from key-worker-recon ");
+        log.info("Starting: Cronjob triggered from key-worker-recon ");
         reconciliationBatchService.reconcileKeyWorkerAllocations();
+        log.info("Complete: Cronjob triggered from key-worker-recon ");
```
neilmendum, line 120: [As discussed on Slack](https://mojdt.slack.com/archives/CP8FR9XL5/p1692183767740989?thread_ts=1692174620.547119&cid=CP8FR9XL5) the alert is triggered if it cannot find a log of `Complete: Key Worker Reconciliation` - your change here, and on the other endpoints, will cause the alert to still fire because the logging is not what it is expecting

We could either
- Update the AppInsights alert to perform a different query
- Update this logging to be the same as before we removed the Camel routes, as an example see [KeyworkerReconRoute.kt](https://github.com/ministryofjustice/keyworker-api/pull/473/files#diff-e8c1ae1cb6921ae6e23543251f0345834b07bc9b78bc6c137a21e26ba7c49defL32)

I would do the second option. Let's chat more if you're unsure

PR #494, opened 2023-09-07
NN-5081 Upgraded spring boot version to 5.4.0 for security fix.

nagendramoj:


build.gradle.kts
```diff
@@ -1,5 +1,5 @@
 plugins {
-  id("uk.gov.justice.hmpps.gradle-spring-boot") version "5.3.0"
+  id("uk.gov.justice.hmpps.gradle-spring-boot") version "5.4.0"
```
neilmendum, line 2: Looks like this, and other gradle deps, are been bumped by https://github.com/ministryofjustice/keyworker-api/pull/489 - we may as well merge that one, and close this one?

For future reference you can bump all dependencies with the gradle task `./gradlew useLatestVersions`

nagendramoj:
Upgrade has been done by renovate, so this PR is redundant

PR #496, opened 2023-09-27
NN-5482 Changed BAD REQUEST to INTERNAL SERVER

nagendramoj:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/ControllerAdvice.java
```diff
@@ -54,7 +54,7 @@ public ResponseEntity<ErrorResponse> handleException(final AccessDeniedException
     @ExceptionHandler(AllocationException.class)
     public ResponseEntity<ErrorResponse> handleException(final AllocationException e) {
```
timoj58, line 55: i think you can leave this as  a 400. its a specific exception (AllocationException)

timoj58 (approved):

PR #504, opened 2023-11-09
SDIT-1174 prison-api CaseNoteResource

steverendell:
Use client id / system tokens when calling these endpoints

### DIFF
--- a/src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/KeyworkerStatsService.java
+++ b/src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/KeyworkerStatsService.java
@@ -103,8 +103,7 @@ public class KeyworkerStatsService {
                     prisonerNosList,
                     range.startDate,
                     range.endDate,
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/KeyworkerStatsService.java
```diff
@@ -103,8 +103,7 @@ public KeyworkerStatsDto getStatsForStaff(final Long staffId, final String priso
                     prisonerNosList,
                     range.startDate,
                     range.endDate,
-                    staffId,
-                    false);
+                    staffId);
```
steverendell, line 106: hardcoding 'admin' to true causes system token to be used

steverendell: the prison-api endoint still has no role protection so far

mikehalmamoj (approved):

PR #509, opened 2023-11-22
NN-5714 Remove role migration

Mjwillis:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/KeyworkerServiceController.kt
```diff
@@ -1160,7 +1158,6 @@ class KeyworkerServiceController(
     }
     if (migrate) {
       keyworkerMigrationService.migrateKeyworkerByPrison(prisonId)
-      roleMigrationService.migrate(prisonId)
```
Mjwillis, line 1163: This doesn't work anyway and we don't need to add roles anymore

timoj58 (approved):

PR #515, opened 2024-01-08
NN-5780 testing transaction change

timoj58:


Mjwillis (approved):
Yes for read only we can leave the transaction off for now

PR #591, opened 2025-01-16
Increase prison api timeout to 60 seconds 

anthony-britton-moj:
as several calls take over 30 seconds to respond successfully


iterative-it (approved):
LGTM

PR #600, opened 2025-02-05
MIA-127

anthony-britton-moj:


helm_deploy/values-prod.yaml
```diff
@@ -43,3 +42,7 @@ generic-prometheus-alerts:
   sqsNumberAlertQueueNames:
     - "hmpps-move-and-improve-production-keyworker_domain_events_dlq"
   sqsAlertsTotalMessagesThreshold: 1
+  applicationCronJobStatusFailedWindowMinutes: 15
+
+schedule:
+  prison_stats: "0 5 * * *"
```
iterative-it, line 48: Small typo here. Prod originally ran at 5 0 * * * so 00:05 rather than 05:00

helm_deploy/values-preprod.yaml
```diff
@@ -29,3 +28,7 @@ generic-prometheus-alerts:
     - "hmpps-move-and-improve-preprod-keyworker_domain_events_dlq"
   sqsAlertsTotalMessagesThreshold: 1
   businessHoursOnly: true
+  applicationCronJobStatusFailedWindowMinutes: 15
```
iterative-it, line 31: Guessing this is what prometheus is alerting us to. What's the default for this?

helm_deploy/values-prod.yaml
```diff
@@ -8,7 +8,6 @@ generic-service:
     AUTH_URI_ROOT: "https://sign-in.hmpps.service.justice.gov.uk/auth"
     SPRING_SECURITY_OAUTH2_RESOURCESERVER_JWT_JWK_SET_URI: "https://sign-in.hmpps.service.justice.gov.uk/auth/.well-known/jwks.json"
     COMPLEXITY_OF_NEED_URI: https://complexity-of-need.hmpps.service.justice.gov.uk
-    PRISONS_WITH_OFFENDERS_THAT_HAVE_COMPLEX_NEEDS: "AGI,BZI,DHI,DWI,ESI,EWI,FHI,LNI,NHI,PFI,SDI,STI"
```
iterative-it, line 11: I can't see anywhere this is still used in the code so probably safe to delete but is it best to keep it in here in case we need to roll back the deployment? Might have misunderstood that point when you made it originally

src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/PrisonKeyworkerConfiguration.kt
```diff
@@ -0,0 +1,14 @@
...
+data class PrisonKeyworkerConfiguration(
+  val isEnabled: Boolean,
+  val hasPrisonersWithHighComplexityNeeds: Boolean,
+  val allowAutoAllocate: Boolean,
+  val capacityTier1: Int,
+  val capacityTier2: Int,
+  val kwSessionFrequencyInWeeks: Int,
+) {
+  companion object {
+    val NOT_CONFIGURED = PrisonKeyworkerConfiguration(false, false, false, 6, 9, 1)
```
iterative-it, line 12: Just noticed that I think the capacityTier1 default used in the migration script in the previous PR was 1 rather than 6. Probably doesn't matter much as it's dev only but it's presumably detectable by them not having been migrated so could be changed to 6 with another migration

src/main/java/uk/gov/justice/digital/hmpps/keyworker/model/PrisonSupported.java
```diff
@@ -56,4 +56,20 @@ public boolean isMigrated() {
     public boolean hasPrisonersWithHighComplexityNeeds() {
         return hasPrisonersWithHighComplexityNeeds;
     }
+
+    public boolean isAutoAllocate() {
```
iterative-it, line 60: Worth a chat on this boolean. I can't immediately see any usage of it in the API. I think it's only used in the UI where it is described as "Allow auto-allocation of prisoners to key workers". It means the buttons for auto allocate appear in the UI. If it is only used in the UI, might be best to rename it fully to `allowAutoAllocation` to show that it's not that it enables some background automatic allocation of people to key workers.

There's actually `private Boolean autoAllocationAllowed` in KeyworkerDto too which similarly looks like it may be a UI only value.

iterative-it (approved):
LGTM

PR #614, opened 2025-02-24
MIA-162 first part with modifications to stats

anthony-britton-moj:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/statistics/PrisonStatisticCalculator.kt
```diff
@@ -151,7 +151,7 @@ class PeopleSummaries(
     personIdentifiers.map {
       PersonSummary(
         it,
-        getReceptionDate(it),
+        getReceptionDate(it).takeIf { LocalDate.now().minusMonths(6).isBefore(it) },
```
iterative-it, line 154: Is this logic something that was in the previous statistics logic?

iterative-it (approved):
LGTM

PR #616, opened 2025-02-25
Limit active keyworkers to those in a given prison

anthony-britton-moj:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/statistics/PrisonStatisticCalculator.kt
```diff
@@ -115,7 +115,7 @@ class PrisonStatisticCalculator(
           ?.let { nomisKeyworkers ->
             val dpsKeyworkers =
               keyworkerRepository
-                .findAll()
+                .findAllKeyworkersForPrison(prisonCode)
```
iterative-it, line 118: Probably worth changing this to findAllActiveKeyworkersForPrison

iterative-it: Ah it's filtering out the inactive allocations rather than inactive keyworkers

iterative-it (approved):
LGTM

PR #629, opened 2025-03-05
MIA-192 listen to case note events

anthony-britton-moj:


src/main/resources/db/migration/V3_0__add_sessions.sql
```diff
@@ -0,0 +1,27 @@
+create table keyworker_session
+(
+    id                uuid       not null primary key,
+    occurred_at       timestamp  not null,
+    person_identifier varchar(7),
+    staff_id          bigint     not null,
```
iterative-it, line 6: We should take the username too. I still want to transition away from staff ids at some point and the usernames from the case notes may help with that

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/SessionAndEntryService.kt
```diff
@@ -0,0 +1,89 @@
...
+    updateKeyworkerInteraction(getCaseNote(personInfo)) {
+      apply {
+        personIdentifier = it.personIdentifier
+      }
+    }
+  }
+
+  fun delete(personInfo: PersonInformation) {
+    when (personInfo.info.subType) {
+      SESSION_SUBTYPE -> ksRepository.deleteById(personInfo.info.id)
```
iterative-it, line 48: Wondering about the value of having separate tables for sessions and entries vs a type of some sort. UR has shown that many users want to remove the KW entry concept completely. Can't see that happening but it does suggest that the direction of travel is fewer KW event types rather than possibly adding a new one

iterative-it: Well argued and I agree

iterative-it (approved):
LGTM

PR #669, opened 2025-04-14
MIA-265 current allocation

anthony-britton-moj:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/GetKeyworkerAllocations.kt
```diff
@@ -5,21 +5,60 @@ import uk.gov.justice.digital.hmpps.keyworker.domain.KeyworkerAllocationReposito
...
+  private val complexityOfNeed: ComplexityOfNeedGateway,
   private val allocationRepository: KeyworkerAllocationRepository,
   private val manageUsers: ManageUsersClient,
   private val staffDetailProvider: StaffDetailProvider,
   private val prisonService: PrisonService,
+  private val caseNotesApiClient: CaseNotesApiClient,
 ) {
-  fun allocationHistoryFor(prisonNumber: String): PersonStaffAllocationHistory {
+  fun currentFor(prisonNumber: String): CurrentPersonStaffAllocation {
+    val level = complexityOfNeed.getOffendersWithMeasuredComplexityOfNeed(setOf(prisonNumber)).firstOrNull()?.level
```
iterative-it, line 33: Since the intended usage of this endpoint is the prisoner profile, it makes sense to reduce the load on complexity of need as much as possible. It should only make this call when the prison is set to have HCN prisoners I reckon. Could be done in the complexity of need service perhaps?

iterative-it (approved):
LGTM

PR #702, opened 2025-05-30
MIA-414 manage staff configuration

anthony-britton-moj:


src/main/resources/db/migration/V10_1__staff_configuration.sql
```diff
@@ -0,0 +1,124 @@
+update reference_data
+set code        = 'STAFF_STATUS_CHANGE',
+    description = 'Staff Status Changed'
+where domain = 'DEALLOCATION_REASON'
+  and code = 'KEYWORKER_STATUS_CHANGE';
+
+update reference_data
+set domain = 'STAFF_STATUS'
```
iterative-it, line 8: This will break the UI for manage KW and changing KW settings temporarily although it's definitely the right thing to do. Might need to coordinate some UI work with this one

iterative-it (approved):
LGTM

PR #710, opened 2025-06-06
MIA-425 switch to one capacity param

anthony-britton-moj:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/PrisonConfigRequest.kt
```diff
@@ -16,6 +15,16 @@ data class PrisonConfigResponse(
...
   val frequencyInWeeks: Int,
-)
+) {
+  companion object {
+    val DEFAULT =
+      PrisonConfigResponse(
+        isEnabled = false,
+        hasPrisonersWithHighComplexityNeeds = false,
+        allowAutoAllocation = false,
+        capacity = 6,
```
iterative-it, line 26: Since this replaces the previous max capacity, should this default be 9 rather than 6?

PR #725, opened 2025-06-16
MIA-441 person search

anthony-britton-moj:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/ReferenceDataController.kt
```diff
@@ -19,7 +19,7 @@ import uk.gov.justice.digital.hmpps.keyworker.services.RetrieveReferenceData
 class ReferenceDataController(
   val referenceData: RetrieveReferenceData,
 ) {
-  @PreAuthorize("hasAnyRole('${Roles.KEYWORKER_RO}', '${Roles.KEYWORKER_RW}')")
+  @PreAuthorize("hasAnyRole('${Roles.KEYWORKER_RO}', '${Roles.KEYWORKER_RW}', '${Roles.ALLOCATIONS_UI}')")
```
iterative-it, line 22: We can remove the other roles from this endpoint thinking about it. UI has the UI role and no one else is expected to call this endpoint

iterative-it (approved):
LGTM

PR #729, opened 2025-06-17
MIA-447 allocation history

anthony-britton-moj:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/GetAllocations.kt
```diff
@@ -0,0 +1,66 @@
...
+  private val allocationRepository: StaffAllocationRepository,
+  private val manageUsers: ManageUsersClient,
+  private val prisonApi: PrisonApiClient,
+  private val prisonService: PrisonService,
+) {
+  fun historyFor(prisonNumber: String): StaffAllocationHistory {
+    val allocations = allocationRepository.findAllByPersonIdentifier(prisonNumber)
+    val usernames = allocations.flatMap { listOfNotNull(it.allocatedBy, it.deallocatedBy) }.toSet()
+    val users = manageUsers.getUsersDetails(usernames).associateBy { it.username }
+    check(users.keys.containsAll(usernames))
```
davidwinchurch, line 26: We've always used SYS as a placeholder for an action performed by the system/non user initiated. Event driven deallocations for KW/PO. It's not a real user in NOMIS so I think this check will fail for any allocation history that contains a deallocation by SYS. Not sure if it's best to have the manage users service return a default for that username or do it here

PR #738, opened 2025-06-26
MIA-516 Staff statistics revamp

bill-moj:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/integration/casenotes/CaseNotesUsage.kt
```diff
@@ -27,31 +27,31 @@ data class UsageByPersonIdentifierRequest(
     fun keyworkerTypes(
       prisonCode: String?,
       personIdentifiers: Set<String>,
-      from: LocalDate,
-      to: LocalDate = from.plusDays(1),
+      from: LocalDateTime,
```
iterative-it, line 30: Do these need to include the time component?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/ReportingPeriod.kt
```diff
@@ -0,0 +1,23 @@
+package uk.gov.justice.digital.hmpps.keyworker.dto
+
+import java.time.LocalDate
+import java.time.LocalDateTime
+
+data class ReportingPeriod(
```
iterative-it, line 6: I like this. Clearer than the from/to params

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/GetStaffDetails.kt
```diff
@@ -77,7 +79,16 @@ class GetStaffDetails(
   fun getDetailsFor(
     prisonCode: String,
     staffId: Long,
+    from: LocalDate?,
+    to: LocalDate?,
   ): StaffDetails {
+    val statsReportingPeriod =
+      if (from != null && to != null) {
+        ReportingPeriod(from.atStartOfDay(), to.plusDays(1).atStartOfDay())
```
iterative-it, line 87: I don't think it should alter the supplied to. If the UI is sending in a range, It's the UI's responsibility to include today if that's the requirement

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/GetStaffDetails.kt
```diff
@@ -116,9 +127,9 @@ class GetStaffDetails(
     val prisonName =
       prisonerDetails.values.firstOrNull()?.prisonName ?: prisonRegisterApi.findPrison(prisonCode)!!.prisonName
 
-    val (current, cnSummary) = allAllocations.staffCountStats(fromDate, now(), prisonConfig, staffId)
+    val (current, cnSummary) = allAllocations.staffCountStats(statsReportingPeriod, prisonConfig, staffId)
     val (previous, _) =
-      allAllocations.staffCountStats(previousFromDate, fromDate, prisonConfig, staffId)
+      allAllocations.staffCountStats(statsReportingPeriod.minusMonths(1), prisonConfig, staffId)
```
iterative-it, line 132: This isn't necessarily the correct decrement any more. If from and to are specified, it should decrement the days between those two rather than a month. Perhaps have a current and previous function in the ReportingPeriod class that calculates this?

Also, with the inclusion of today in the current, will there be an overlap with the end day of previous and the first day of current? I think the ranges used for today would be:

current: 27/05/2025 00:00 -> 28/06/2025 00:00
previous: 27/04/2025 00:00 -> 28/05/2025 00:00

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/GetStaffDetails.kt
```diff
@@ -193,31 +203,26 @@ class GetStaffDetails(
...
-      if (context.policy == AllocationPolicy.KEY_WORKER) {
-        applicableAllocations.sumOf { it.daysAllocatedForStats(from, to) }
-      } else {
-        0
-      }
-    val averagePerDay = if (total == 0L) 0 else total / DAYS.between(from, to)
     val projectedSessions =
-      if (averagePerDay == 0L) {
-        0
+      if (context.policy == AllocationPolicy.KEY_WORKER && count() > 0) {
```
iterative-it, line 219: Guessing you'll be doing personal officer in a later PR?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/integration/casenotes/CaseNotesUsage.kt
bill-moj: this function generate a UsageByPersonIdentifierRequest which we want to have the exact DateTime provided so it's better to pass to datetime to it?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/GetStaffDetails.kt
bill-moj: projected session should have the same calculation for both policy?
i suppose i can remove the `if` and update the tests accordingly

src/main/java/uk/gov/justice/digital/hmpps/keyworker/statistics/PrisonStatisticCalculator.kt
```diff
@@ -74,7 +74,7 @@ class PrisonStatisticCalculator(
           .associateBy { it.personIdentifier }
 
       val cnSummary =
-        caseNotesApi.getUsageByPersonIdentifier(keyworkerTypes(prisonCode, eligiblePrisoners, date)).summary()
+        caseNotesApi.getUsageByPersonIdentifier(keyworkerTypes(prisonCode, eligiblePrisoners, date.atStartOfDay())).summary()
```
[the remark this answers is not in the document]
bill-moj: functionally no change is made here.
`caseNotesApi.getUsageByPersonIdentifier` has always required DateTime in its request.
the time component was previously appended by the helper function `keyworkerTypes`.

since we supply the exact DateTime in `GetStaffDetails`, we don't need the helper function to append the time component. side effect is that we have to append the time component here where the same helper function is used.

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/GetStaffDetails.kt
iterative-it: Ah no it's great that you have them as optional params. What I mean is when to is supplied i.e. `to != null`, you're still adding one day to the supplied value. That's the bit I felt was wrong. Wouldn't it be better to use the supplied value as is?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/integration/casenotes/CaseNotesUsage.kt
iterative-it: Right I see. I can't see anywhere we use a date time so maybe the correct thing to do is change the UsageByPersonIdentifierRequest from and to to a local date instead?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/GetStaffDetails.kt
bill-moj: when user input for example 2025/05/01 to 2025/05/31, they would expect it to cover all events up to 2025/05/31 23:59:59.

i think we converts 2025/05/31 to 2025/06/01 00:00 only so that the reporting period will actually cover the whole of the previous day. the plus 1 day logic seems more like an api inner logic thing than a UI responsibility from my point of view.

iterative-it: I don't agree but I'm not going to argue it further since we don't have any requirements yet to supply this from the front end 😄

iterative-it (approved):
LGTM

PR #746, opened 2025-07-02
MIA-517 patch staff config+jobclassification endpoint

bill-moj:
MIA-517

### DIFF
--- a/src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/StaffConfigManager.kt
+++ b/src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/StaffConfigManager.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.keyworker.services
 
+import jakarta.persistence.EntityNotFoundException
 import org.springframework.stereotype.Service
 import org.springframework.transaction.annotation.Transactional
 import uk.gov.justice.digital.hmpps.keyworker.config.AllocationContext
@@ -14,11 +15,15 @@ import uk.gov.justice.digital.hmpps.keyworker.domain.StaffRoleRepository
 import uk.gov.justice.digital.hmpps.keyworker.domain.getKeyworkerStatus
 import uk.gov.justice.digital.hmpps.keyworker.domain.getReferenceData
 import uk.gov.justice.digital.hmpps.keyworker.domain.of
+import uk.gov.justice.digital.hmpps.keyworker.dto.PatchStaffConfigRequest
+import uk.gov.justice.digital.hmpps.keyworker.dto.PatchStaffJobClassificationRequest
 import uk.gov.justice.digital.hmpps.keyworker.dto.StaffConfigRequest
+import uk.gov.justice.digital.hmpps.keyworker.integration.PrisonApiClient
 import uk.gov.justice.digital.hmpps.keyworker.integration.nomisuserroles.JobClassification
 import uk.gov.justice.digital.hmpps.keyworker.integration.nomisuserroles.NomisUserRolesApiClient
 import uk.gov.justice.digital.hmpps.keyworker.integration.nomisuserroles.StaffJobClassificationRequest
 import uk.gov.justice.digital.hmpps.keyworker.model.DeallocationReason.STAFF_STATUS_CHANGE
+import java.time.LocalDate
 
 @Transactional
 @Service
@@ -28,6 +33,7 @@ class StaffConfigManager(
   private val allocationRepository: StaffAllocationRepository,
   private val nurApi: NomisUserRolesApiClient,
   private val staffRoleRepository: StaffRoleRepository,
+  private val prisonApi: PrisonApiClient,
 ) {
   fun setStaffConfiguration(
     prisonCode: String,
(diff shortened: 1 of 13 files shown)


src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/StaffConfigManager.kt
```diff
@@ -58,6 +64,86 @@ class StaffConfigManager(
     }
   }
 
+  fun patchStaffConfiguration(
+    prisonCode: String,
+    staffId: Long,
+    request: PatchStaffConfigRequest,
+  ) {
+    val staffConfig = staffConfigRepository.findByStaffId(staffId) ?: throw EntityNotFoundException("Staff config not found")
```
iterative-it, line 72: I think this may not cover the staff config ACs properly, particularly 1c. It shouldn't throw an exception if it's not found. Instead if there is no config, it should check if any of status, capacity or reactivateOn are non null and create the relevant config instead.

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/StaffConfigManager.kt
```diff
@@ -58,6 +64,86 @@ class StaffConfigManager(
...
+    if (!request.jobClassification.isPresent || request.jobClassification.get() != null) {
+      staffConfig.patch(request)
+    }
+
+    if (request.deactivateActiveAllocations.orElse(false) ||
```
iterative-it, line 77: I'd switch the order if this deactivate and the removal of the job classification. We have to call another API to remove the KW job responsibility which could fail leaving the staff member with no allocations but still being a key worker

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/StaffConfigManager.kt
```diff
@@ -58,6 +64,86 @@ class StaffConfigManager(
...
+        it,
+        StaffJobClassificationRequest(
+          position = position,
+          scheduleType = scheduleType,
+          hoursPerWeek = hoursPerWeek,
+          fromDate = fromDate,
+          toDate = LocalDate.now(),
+        ),
+      )
+    } ?: throw EntityNotFoundException("Staff role not found")
```
iterative-it, line 110: For idempotency purposes, it's fine if it can't find the existing job classification. Means it's already deactivated. This shouldn't throw an exception

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/StaffConfigManager.kt
```diff
@@ -58,6 +64,86 @@ class StaffConfigManager(
...
+  } ?: run {
+    staffRoleRepository.findByPrisonCodeAndStaffId(prisonCode, staffId)?.apply {
+      request.position.ifPresent { position = referenceDataRepository.getReferenceData(ReferenceDataDomain.STAFF_POSITION of it) }
+      request.scheduleType.ifPresent {
+        scheduleType =
+          referenceDataRepository.getReferenceData(ReferenceDataDomain.STAFF_SCHEDULE_TYPE of it)
+      }
+      request.hoursPerWeek.ifPresent { hoursPerWeek = it }
+      request.fromDate.ifPresent { fromDate = it }
+    } ?: throw EntityNotFoundException("Staff role not found")
```
iterative-it, line 144: Same as above. Needs to create rather than throw

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/StaffConfigManager.kt
```diff
@@ -58,6 +69,138 @@ class StaffConfigManager(
...
 
+  fun setStaffDetails(
+    prisonCode: String,
+    staffId: Long,
+    request: StaffDetailsRequest,
+  ) {
+    val policy = AllocationContext.get().policy
+    val prisonConfig = prisonConfigRepository.findByCode(prisonCode) ?: PrisonConfiguration.default(prisonCode)
+    val staffConfig = staffConfigRepository.findByStaffId(staffId)
+    val staffRole: Pair<StaffRole?, NomisStaffRole?>? =
```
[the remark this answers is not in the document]
iterative-it: The use of !!.first!!, !!.second!! later on in particular

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/StaffConfigManager.kt
```diff
@@ -58,6 +69,138 @@ class StaffConfigManager(
...
+            reactivateOn.orElse(null),
+            staffId,
+          ),
+        )
+      }
+    } else if (!deactivateJobRoleAndConfig) {
+      staffConfig.patch(request)
+    }
+
+    if (removeActiveAllocations) {
```
iterative-it, line 123: Similar to the previous implementation, I think it would be better to deallocate last in case the call to the NOMIS User and Roles API fails

iterative-it (approved):
LGTM

PR #757, opened 2025-07-10
Exclude records with eligibility date older than 6 months

anthony-britton-moj:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/statistics/PrisonStatisticCalculator.kt
```diff
@@ -126,8 +126,8 @@ class PrisonStatisticCalculator(
       val overSixMonths =
         summaries.data.filter {
```
iterative-it, line 127: I think the > 6 months ones will have been filtered out by the `eligibilityDateIsValid()` check when creating the summaries so we won't get any logged. Could add `eligibilityDateIsValid` as a property of `PersonSummary` and filter it in `PeopleSummaries` `averageDaysToAllocation` and `averageDaysToSession`. That way we can trial improvements to the two stats by logging the possible anomalies but not store them

PR #758, opened 2025-07-11
MIA-508 current allocations

anthony-britton-moj:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/controllers/CurrentAllocationController.kt
```diff
@@ -0,0 +1,22 @@
...
+import uk.gov.justice.digital.hmpps.keyworker.dto.CurrentPersonStaffAllocation
+import uk.gov.justice.digital.hmpps.keyworker.services.GetCurrentAllocations
+
+@RestController
+class CurrentAllocationController(
+  private val allocations: GetCurrentAllocations,
+) {
+  @Tag(name = MANAGE_ALLOCATIONS)
+  @PreAuthorize("hasRole('${Roles.ALLOCATIONS_RO}')")
+  @GetMapping("/prisoners/{personIdentifier}/allocations/current")
```
iterative-it, line 18: Ah totally missed this in the ACs. This will need adding to the `SentryConfig.isHighUsage` function so we don't overuse our quota

PR #765, opened 2025-07-18
MIA-626

anthony-britton-moj:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/StaffConfigManager.kt
```diff
@@ -45,131 +50,125 @@ class StaffConfigManager(
...
-        }
-      } ?: run {
-        staffRole!!.first!!.apply {
-          request.staffRole.get()!!.let { jobRoleRequest ->
-            jobRoleRequest.position.ifPresent {
-              position =
-                referenceDataRepository.getReferenceData(ReferenceDataDomain.STAFF_POSITION of it)
+    when (roleAction) {
+      null -> return
+      Action.CREATE -> {
```
iterative-it, line 77: Could maybe break these action implementations out into separate functions to make this function a bit easier to read but it's already a big improvement

iterative-it (approved):
LGTM

PR #788, opened 2025-08-01
MIA-653

anthony-britton-moj:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/DeallocationService.kt
```diff
@@ -0,0 +1,29 @@
...
+) {
+  fun deallocateExpiredAllocations(
+    prisonCode: String,
+    personIdentifier: String,
+    deallocationReason: DeallocationReason,
+  ) {
+    val reason = requireNotNull(referenceDataRepository.findByKey(DEALLOCATION_REASON of deallocationReason.reasonCode))
+    val allocations =
+      allocationRepository.findActiveForAllPolicies(personIdentifier).mapNotNull {
+        it.takeIf { it.prisonCode != prisonCode }?.apply { deallocate(reason) }
```
iterative-it, line 25: There used to be a check that the destination prison code was a prison (`nomisService.isPrison(`) which means this would deallocate on transfers to hospitals, immigration centres etc. I think that's fine as a standard external movement to a TAP destination or court wouldn't be a transfer, release or arrival. Would you agree it's safe?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/events/OffenderEventListener.kt
```diff
@@ -5,42 +5,60 @@ import com.fasterxml.jackson.module.kotlin.readValue
...
+  companion object {
+    const val EXTERNAL_MOVEMENT = "EXTERNAL_MOVEMENT_RECORD-INSERTED"
+  }
 }
+
+data class OffenderEvent(
+  val bookingId: Long? = null,
+  val movementSeq: Long? = null,
+  val offenderIdDisplay: String? = null,
+  val movementDateTime: LocalDateTime? = null,
```
iterative-it, line 57: Do you think we should use the `movementDateTime` for deallocated at? Would cope with delayed or requeued DLQ messages better

iterative-it (approved):
LGTM

PR #789, opened 2025-08-05
MIA-387

anthony-britton-moj:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/ReactivateStaff.kt
```diff
@@ -0,0 +1,33 @@
...
+
+@Transactional
+@Service
+class ReactivateStaff(
+  private val staffConfigRepository: StaffConfigRepository,
+  private val referenceDataRepository: ReferenceDataRepository,
+) {
+  fun returningFromLeave(date: LocalDate) {
+    staffConfigRepository
+      .findAllStaffReturningFromLeave(date.plusDays(1))
```
iterative-it, line 21: Could maybe change the SQL to use `<= :date` then only send date rather than `date.plusDays(1)`?

src/test/java/uk/gov/justice/digital/hmpps/keyworker/integration/ReturningFromLeaveIntTest.kt
```diff
@@ -0,0 +1,80 @@
...
+import uk.gov.justice.digital.hmpps.keyworker.config.AllocationContext.Companion.SYSTEM_USERNAME
+import uk.gov.justice.digital.hmpps.keyworker.config.AllocationPolicy
+import uk.gov.justice.digital.hmpps.keyworker.domain.StaffConfiguration
+import uk.gov.justice.digital.hmpps.keyworker.model.StaffStatus
+import java.time.LocalDate
+
+class ReturningFromLeaveIntTest : IntegrationTest() {
+  @ParameterizedTest
+  @EnumSource(AllocationPolicy::class)
+  fun `returning from leave does update due to return tomorrow`(policy: AllocationPolicy) {
```
iterative-it, line 18: I think this should be "returning from leave does **not** update due to return tomorrow"

iterative-it (approved):
LGTM

PR #799, opened 2025-08-11
MIA-438 PO migration

anthony-britton-moj:


src/main/resources/db/migration/V13_4__personal_officer_history.sql
```diff
@@ -0,0 +1,88 @@
...
+           allocated_by,
+           deallocated_at,
+           deallocated_by,
+           not is_active
+    into _allocated_at, _allocated_by, _deallocation_at, _deallocated_by, _deallocated
+    from allocation
+    where id = allocation_id;
+
+    _rev_id = nextval('audit_revision_id_seq');
+    insert into audit_revision(id, timestamp, username, caseload_id, affected_entities)
```
iterative-it, line 22: I'm assuming this will be run after clearing the existing audit history? Worth adding a check for existing rows or a delete of those rows anyway to stop duplicate rows?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/migration/Movement.kt
```diff
@@ -9,8 +9,8 @@ data class Movement(
   val offenderNo: String,
   val movementDate: LocalDate,
   val movementTime: LocalTime,
-  val fromAgency: String,
-  val toAgency: String,
+  val fromAgency: String?,
+  val toAgency: String?,
```
iterative-it, line 13: Can't imagine this would ever be null for `ADM` movements but if it was, I guess it would match the `earliestArrivalAtAnotherPrison` check. Null is a different prison though so likely fine anyway!

iterative-it (approved):
LGTM

PR #806, opened 2025-08-12
MIA-685 add comparison dates for stats endpoints

anthony-britton-moj:


src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/GetStaffDetails.kt
```diff
@@ -82,14 +82,12 @@ class GetStaffDetails(
...
+    comparisonTo: LocalDate?,
     includeStats: Boolean,
   ): StaffDetails {
     val reportingPeriod =
-      if (from != null && to != null) {
-        ReportingPeriod(from.atStartOfDay(), to.atStartOfDay())
-      } else {
-        ReportingPeriod.currentMonth()
-      }
+      ReportingPeriod.of(from, to, ReportingPeriod.of(comparisonFrom, comparisonTo)) ?: ReportingPeriod.currentMonth()
```
iterative-it, line 90: Should the null coalesce be `ReportingPeriod. previousPeriod()`?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/dto/ReportingPeriod.kt
```diff
@@ -7,18 +7,31 @@ import java.time.temporal.ChronoUnit.DAYS
...
   companion object {
+    fun of(
+      from: LocalDate?,
+      to: LocalDate?,
+      previous: ReportingPeriod? = null,
+    ): ReportingPeriod? =
+      if (from == null || to == null) {
+        null
+      } else {
+        ReportingPeriod(from.atStartOfDay(), to.atStartOfDay())
```
iterative-it, line 32: Previous isn't passed in here?

src/main/java/uk/gov/justice/digital/hmpps/keyworker/services/PrisonStatsService.kt
```diff
@@ -24,10 +24,12 @@ class PrisonStatsService(
     prisonCode: String,
     from: LocalDate,
     to: LocalDate,
+    comparisonFrom: LocalDate?,
+    comparisonTo: LocalDate?,
   ): PrisonStats {
     val dateRange = DAYS.between(from, to.plusDays(1))
-    val previousTo = from.minusDays(1)
-    val previousFrom = from.minusDays(dateRange)
+    val previousTo = comparisonTo ?: from.minusDays(1)
```
iterative-it, line 31: If either comparisonTo or comparisonFrom are null, both previousTo and previousFrom should be calculated. This will use combinations of supplied comparison dates and initial dates

iterative-it: Yeah that's true but the staff path does account for it. Also one uses ReportingPeriod and the other doesn't. Neither is an issue so it's up to you if you want to change it or do any form of a refactor to make these ranges uses common code

iterative-it (approved):
LGTM

PR #866, opened 2025-09-15
MIA-845: Set establishment sort order

sameer-moj:


src/main/resources/db/migration/V15_8__keyworker_allocation_ordering.sql
```diff
@@ -0,0 +1,23 @@
+ALTER TABLE prison_configuration ADD COLUMN IF NOT EXISTS allocation_order VARCHAR (15);
+ALTER TABLE prison_configuration ALTER COLUMN allocation_order SET DEFAULT 'BY_ALLOCATIONS';
+
+UPDATE prison_configuration SET allocation_order = 'BY_ALLOCATIONS' WHERE allocation_order IS NULL;
```
bill-moj, line 4: probably don't need `SET DEFAULT 'BY_ALLOCATIONS'`

iterative-it: We'd also need to populate the existing rows. This is OK but I've made a note to remove this default when we remove all the legacy code.

There are a few changes needed here though. You should be able to combine all of this into one line like `ALTER TABLE prison_configuration ADD COLUMN IF NOT EXISTS allocation_order VARCHAR (15) NOT NULL DEFAULT 'BY_ALLOCATIONS';`

src/main/java/uk/gov/justice/digital/hmpps/keyworker/domain/PrisonConfiguration.kt
```diff
@@ -34,6 +36,9 @@ class PrisonConfiguration(
   var frequencyInWeeks: Int,
   @Column(name = "has_prisoners_with_high_complexity_needs")
   var hasPrisonersWithHighComplexityNeeds: Boolean,
+  @Enumerated(EnumType.STRING)
+  @Column(name = "allocation_order", nullable = false)
```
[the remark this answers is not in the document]
sameer-moj: Remvoed

iterative-it (approved):
LGTM

PR #958, opened 2026-04-23
LSM-442: Increase Limits

theDustRoom:
Increase keyworker-api memory limits based on observed ~1.1Gi pod usage


eve-bramley (approved):
LGTM

PR #959, opened 2026-04-29
Syncing who is a key worker from NOMIS

cugorji-justice:
https://dsdmoj.atlassian.net/browse/LSM-444

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/keyworker/domain/StaffRole.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/keyworker/domain/StaffRole.kt
@@ -63,6 +63,11 @@ interface StaffRoleRepository : JpaRepository<StaffRole, UUID> {
 
   fun findAllByPrisonCode(prisonCode: String): List<StaffRole>
 
+  fun findAllByPrisonCodeAndPolicy(
+    prisonCode: String,
+    policy: String,
+  ): List<StaffRole>
+
   @Query(
     """
         select * from staff_role where prison_code = :prisonCode and staff_id = :staffId and policy_code = :policyCode
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/keyworker/services/staff/KeyworkerStaffRoleSync.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/keyworker/services/staff/KeyworkerStaffRoleSync.kt
@@ -1,0 +1,56 @@
+package uk.gov.justice.digital.hmpps.keyworker.services.staff
+
+import org.springframework.stereotype.Service
+import org.springframework.transaction.support.TransactionTemplate
+import uk.gov.justice.digital.hmpps.keyworker.config.AllocationContext
+import uk.gov.justice.digital.hmpps.keyworker.config.AllocationPolicy
+import uk.gov.justice.digital.hmpps.keyworker.config.set
+import uk.gov.justice.digital.hmpps.keyworker.domain.StaffRoleRepository
+import uk.gov.justice.digital.hmpps.keyworker.integration.nomisuserroles.StaffJobClassificationRequest
+import uk.gov.justice.digital.hmpps.keyworker.integration.prisonapi.NomisStaffRole
+import uk.gov.justice.digital.hmpps.keyworker.integration.prisonapi.PrisonApiClient
+import uk.gov.justice.digital.hmpps.keyworker.services.PrisonService
+import java.time.LocalDate
+
+@Service
+class KeyworkerStaffRoleSync(
+  private val prisonService: PrisonService,
+  private val prisonApi: PrisonApiClient,
+  private val staffConfigManager: StaffConfigManager,
+  private val staffRoleRepository: StaffRoleRepository,
+  private val transactionTemplate: TransactionTemplate,
+) {
+  fun syncKeyworkerStaffRoles() {
+    val originalContext = AllocationContext.get()
+
+    try {
+      originalContext.copy(policy = AllocationPolicy.KEY_WORKER).set()
+      prisonService.findPolicyEnabledPrisons(AllocationPolicy.KEY_WORKER.name).forEach(::syncKeyworkerStaffRoles)
+    } finally {
+      originalContext.set()
+    }
+  }
+
+  fun syncKeyworkerStaffRoles(prisonCode: String) {
+    val keyworkers = prisonApi.getKeyworkersForPrison(prisonCode)
+    val originalContext = AllocationContext.get()
+
+    try {
+      transactionTemplate.execute {
+        originalContext.copy(activeCaseloadId = prisonCode, policy = AllocationPolicy.KEY_WORKER).set()
+        val activeKeyworkerStaffIds = keyworkers.map { it.staffId }.toSet()
+        keyworkers.forEach {
+          staffConfigManager.setStaffRole(AllocationPolicy.KEY_WORKER, prisonCode, it.staffId, it.asRequest())
+        }
+        staffRoleRepository
+          .findAllByPrisonCodeAndPolicy(prisonCode, AllocationPolicy.KEY_WORKER.name)
+          .filterNot { it.staffId in activeKeyworkerStaffIds }
+          .forEach { it.toDate = LocalDate.now() }
+      }
+    } finally {
+      originalContext.set()
+    }
+  }
+
+  private fun NomisStaffRole.asRequest() = StaffJobClassificationRequest(position, scheduleType, hoursPerWeek, fromDate, toDate)
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/keyworker/services/staff/NomisStaffRoleUpdateEvent.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/keyworker/services/staff/NomisStaffRoleUpdateEvent.kt
@@ -1,0 +1,35 @@
+package uk.gov.justice.digital.hmpps.keyworker.services.staff
+
+import org.slf4j.LoggerFactory
+import org.springframework.scheduling.annotation.Async
+import org.springframework.stereotype.Component
+import org.springframework.transaction.event.TransactionPhase.AFTER_COMMIT
+import org.springframework.transaction.event.TransactionalEventListener
+import uk.gov.justice.digital.hmpps.keyworker.integration.nomisuserroles.NomisUserRolesApiClient
+import uk.gov.justice.digital.hmpps.keyworker.integration.nomisuserroles.StaffJobClassificationRequest
+
+data class NomisStaffRoleUpdateEvent(
+  val prisonCode: String,
+  val staffId: Long,
+  val role: String,
+  val request: StaffJobClassificationRequest,
+)
+
+@Component
+class NomisStaffRoleUpdateListener(
+  private val nomisUserRolesApiClient: NomisUserRolesApiClient,
+) {
+  @Async
+  @TransactionalEventListener(phase = AFTER_COMMIT)
+  fun updateNomisStaffRole(event: NomisStaffRoleUpdateEvent) {
+    runCatching {
+      nomisUserRolesApiClient.setStaffRole(event.prisonCode, event.staffId, event.role, event.request)
+    }.onFailure {
+      log.warn("Failed to update NOMIS staff role {} for staff {} at {}", event.role, event.staffId, event.prisonCode, it)
+    }
+  }
+
+  companion object {
+    private val log = LoggerFactory.getLogger(this::class.java)
+  }
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/keyworker/services/staff/StaffConfigManager.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/keyworker/services/staff/StaffConfigManager.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.keyworker.services.staff
 
+import org.springframework.context.ApplicationEventPublisher
 import org.springframework.stereotype.Service
 import org.springframework.transaction.annotation.Transactional
 import uk.gov.justice.digital.hmpps.keyworker.config.AllocationContext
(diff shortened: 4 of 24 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/keyworker/domain/StaffRole.kt
```diff
@@ -63,6 +63,11 @@ interface StaffRoleRepository : JpaRepository<StaffRole, UUID> {
 
   fun findAllByPrisonCode(prisonCode: String): List<StaffRole>
 
+  fun findAllByPrisonCodeAndPolicy(
+    prisonCode: String,
+    policy: String,
```
anthony-britton-moj, line 68: This isn't actually required here. The above function of `findAllByPrisonCode` already does this. There is a `PolicyResolver` that will set the policy based on the context automatically - and setting it in the query will actually make no difference. The policy resolver automatically hides the rows from other policies - regardless of the value passed in.

src/main/kotlin/uk/gov/justice/digital/hmpps/keyworker/services/staff/KeyworkerStaffRoleSync.kt
```diff
@@ -0,0 +1,56 @@
...
+  private val transactionTemplate: TransactionTemplate,
+) {
+  fun syncKeyworkerStaffRoles() {
+    val originalContext = AllocationContext.get()
+
+    try {
+      originalContext.copy(policy = AllocationPolicy.KEY_WORKER).set()
+      prisonService.findPolicyEnabledPrisons(AllocationPolicy.KEY_WORKER.name).forEach(::syncKeyworkerStaffRoles)
+    } finally {
+      originalContext.set()
```
anthony-britton-moj, line 30: Given this is fired from a http request - the request config `KeyworkerContextInterceptor` should set and unset the context automatically. 

Looking at the code - it looks like the function `override fun postHandle` should actually be `override fun afterCompletion`. Doing that would unset using the AllocationContext.clear() function rather than resetting it to a given value. These request contexts are not designed to be reused so should just be cleared afterwards rather than reset to a previous value.

If you update to use the afterCompletion you won't need the try and finally block and it will keep the code cleaner and easier to read for future devs.

src/main/kotlin/uk/gov/justice/digital/hmpps/keyworker/services/staff/KeyworkerStaffRoleSync.kt
```diff
@@ -0,0 +1,56 @@
...
+        val activeKeyworkerStaffIds = keyworkers.map { it.staffId }.toSet()
+        keyworkers.forEach {
+          staffConfigManager.setStaffRole(AllocationPolicy.KEY_WORKER, prisonCode, it.staffId, it.asRequest())
+        }
+        staffRoleRepository
+          .findAllByPrisonCodeAndPolicy(prisonCode, AllocationPolicy.KEY_WORKER.name)
+          .filterNot { it.staffId in activeKeyworkerStaffIds }
+          .forEach { it.toDate = LocalDate.now() }
+      }
+    } finally {
```
anthony-britton-moj, line 50: Same comment about the context as above. I'd remove the try/finally and use the context interceptor with the after complete to do this automatically.

src/main/kotlin/uk/gov/justice/digital/hmpps/keyworker/services/staff/KeyworkerStaffRoleSync.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    try {
+      transactionTemplate.execute {
+        originalContext.copy(activeCaseloadId = prisonCode, policy = AllocationPolicy.KEY_WORKER).set()
+        val activeKeyworkerStaffIds = keyworkers.map { it.staffId }.toSet()
+        keyworkers.forEach {
+          staffConfigManager.setStaffRole(AllocationPolicy.KEY_WORKER, prisonCode, it.staffId, it.asRequest())
+        }
+        staffRoleRepository
+          .findAllByPrisonCodeAndPolicy(prisonCode, AllocationPolicy.KEY_WORKER.name)
```
anthony-britton-moj, line 46: As you've set the AllocationPolicy in the AllocationContext - this function isn't required - just use the normal `findAllByPrisonCode`. The PolicyResolver will auto-filter rows based on the correct policy.

src/main/kotlin/uk/gov/justice/digital/hmpps/keyworker/services/staff/NomisStaffRoleUpdateEvent.kt
```diff
@@ -0,0 +1,35 @@
...
+  val role: String,
+  val request: StaffJobClassificationRequest,
+)
+
+@Component
+class NomisStaffRoleUpdateListener(
+  private val nomisUserRolesApiClient: NomisUserRolesApiClient,
+) {
+  @Async
+  @TransactionalEventListener(phase = AFTER_COMMIT)
```
anthony-britton-moj, line 23: Nicely done, making use of the async and transaction. 
For the failure case you might want to record this to sentry. Since you are catching the exception here and it's in an async context - consider `Sentry.Sentry.captureException(ex)` - then you would get an alert in your slack channel if this happens rather than relying on finding it in the logs.

src/main/kotlin/uk/gov/justice/digital/hmpps/keyworker/services/staff/StaffConfigManager.kt
```diff
@@ -184,9 +174,21 @@ class StaffConfigManager(
...
+    staffId: Long,
+    request: StaffJobClassificationRequest,
+  ) {
+    policy.nomisUserRoleCode?.also {
+      applicationEventPublisher.publishEvent(NomisStaffRoleUpdateEvent(prisonCode, staffId, it, request))
+    }
+  }
 
   private fun JobClassification.asStaffRole(
+    policy: AllocationPolicy,
```
anthony-britton-moj, line 191: You don't need to manually set the policy here -this is set from the context as the policy is required.

src/main/kotlin/uk/gov/justice/digital/hmpps/keyworker/services/staff/StaffRoleRetriever.kt
```diff
@@ -39,86 +34,28 @@ data class StaffSummaryWithRole(
...
 
   override fun getActivePoliciesForPrison(
     prisonCode: String,
     staffId: Long,
   ): Set<AllocationPolicy> = staffRoleRepository.findActiveStaffPoliciesForPrison(prisonCode, staffId)
 
   override fun getStaffRoles(prisonCode: String): Map<Long, StaffRoleInfo> =
-    staffRoleRepository.findAllByPrisonCode(prisonCode).associate { it.staffId to it.roleInfo() }
+    staffRoleRepository
+      .findAllByPrisonCodeAndPolicy(prisonCode, AllocationContext.get().requiredPolicy().name)
```
anthony-britton-moj, line 51: similar comment around not setting the policy manually here - this is done automatically

src/main/kotlin/uk/gov/justice/digital/hmpps/keyworker/services/staff/StaffRoleRetriever.kt
```diff
@@ -39,86 +34,28 @@ data class StaffSummaryWithRole(
...
+      .associate { it.staffId to it.roleInfo() }
 
   override fun getStaffWithRoles(prisonCode: String): List<StaffSummaryWithRole> {
-    val roles = staffRoleRepository.findAllByPrisonCode(prisonCode).associate { it.staffId to it.roleInfo() }
+    val roles =
+      staffRoleRepository
+        .findAllByPrisonCodeAndPolicy(prisonCode, AllocationContext.get().requiredPolicy().name)
```
anthony-britton-moj, line 57: similar comment around not setting the policy manually here - this is done automatically

src/main/kotlin/uk/gov/justice/digital/hmpps/keyworker/services/staff/StaffRoleRetriever.kt
```diff
@@ -130,7 +67,9 @@ class LocalRoleRetriever(
     prisonApi.getStaffSummariesFromIds(setOf(staffId)).firstOrNull()?.let {
       StaffSummaryWithRole(
         it,
-        staffRoleRepository.findByPrisonCodeAndStaffId(prisonCode, staffId)?.toModel(),
+        staffRoleRepository
+          .findRoleIncludingInactiveForPolicy(prisonCode, staffId, AllocationContext.get().requiredPolicy().name)
```
anthony-britton-moj, line 71: Similar thing here - you've switched to a native query - designed for a use case when multiple policies are required at the same time. In this case, it seems you only need a single context so I would stick with the auto policy selection and the original `staffRoleRepository.findByPrisonCodeAndStaffId`

src/test/kotlin/uk/gov/justice/digital/hmpps/keyworker/integration/AllocatableStaffSearchIntegrationTest.kt
```diff
@@ -447,6 +416,9 @@ class AllocatableStaffSearchIntegrationTest : IntegrationTest() {
 
   companion object {
     const val SEARCH_URL = "/search/prisons/{prisonCode}/staff-allocations"
+    private val STAFF_ID_OFFSET = (System.currentTimeMillis() % 100_000L) * 1_000L
+
+    private fun newStaffId() = STAFF_ID_OFFSET + newId()
```
anthony-britton-moj, line 421: Was there a problem with id generation here? Not sure why this has been introduced? Not that it's an issue, just trying to understand what this is resolving.

src/test/kotlin/uk/gov/justice/digital/hmpps/keyworker/integration/CalculatePrisonStatisticsTest.kt
```diff
@@ -286,6 +282,8 @@ class CalculatePrisonStatisticsTest : IntegrationTest() {
       },
     )
   }
+
+  private fun uniquePrisonCode(prefix: String) = "$prefix${newId() % 10_000L}"
```
anthony-britton-moj, line 286: if you want unique prison codes - I'd be careful with this version as a prison code has a limited number of characters and this requires any dev to know/understand all the limitations eg only using x chars so you don't exceed the limit. It also might resuse numbers if the new id value is called sufficiently - meaning they are no longer unique. Easier to have a function that takes care of that and gives more realistic looking data.

In other projects we have used something like
`fun prisonCode(attempts: Int = 10): String {
    if (attempts <= 0) throw IllegalStateException("Ran out of attempts to find a unique prison code")
    val prisonCode = (1..3).map { letters.random() }.joinToString("")
    return if (usedPrisonCodes.add(prisonCode)) {
      prisonCode
    } else {
      prisonCode(attempts - 1)
    }
  }`

src/test/kotlin/uk/gov/justice/digital/hmpps/keyworker/integration/ManageStaffDetailsIntTest.kt
anthony-britton-moj, line 222: This part of the test probably needs to run for all policies now - then you wouldn't need to update below where the logic for checking AllocationPolicy is required as the role is being generated now as part of the config update. I would just remove the else block

src/test/kotlin/uk/gov/justice/digital/hmpps/keyworker/integration/ManageStaffDetailsIntTest.kt
```diff
@@ -318,7 +325,11 @@ class ManageStaffDetailsIntTest : IntegrationTest() {
       staffConfig,
       staffConfig.id,
       RevisionType.MOD,
-      setOf(StaffConfiguration::class.simpleName!!),
+      if (policy == AllocationPolicy.KEY_WORKER) {
```
anthony-britton-moj, line 328: Similar pattern here as the comment above. Now all policies are being generated and stored in the db - you probably don't need the if keyworker checks to do different things in the tests - you want it to work the same regardless of policy. Possibly worth checking the other tests you've updated - and having a consistent approach throughout them all that they work the same regardless of policy

anthony-britton-moj:
Overall this seems to cover everything that was in the ticket. Nice one.

One thing to be aware of here addition to any of the other comments. The original ticket was intended to be deployed in stages. Since you've done all the work in one go - that means at the point of deployment this code will start reading from the databsae for staff role information for keyworkers. This data will not be present until after the sync job has ran for each prison to populate their keyworkers. 

I would probably have done the loading into the db as a separate PR from that of reading from the DB. Then you could leave the API call in place for the reads while you verify the db has the data populated. Then deploy the read from db part. Otherwise, you may need to run this out of hours - or at least make someone aware that staff role information will not be available for keyworkers for a period of time. In theory it won't be very long - perhaps testing in preprod will show that it's negligible - just something to be aware of.

One other thing - if you want to run in parallel and reduce the impact of time you can use the `PrisonStatisticsTrigger` as an example of running the job for each prison from a queue.

src/main/kotlin/uk/gov/justice/digital/hmpps/keyworker/services/staff/KeyworkerStaffRoleSync.kt
```diff
@@ -0,0 +1,56 @@
...
+          .filterNot { it.staffId in activeKeyworkerStaffIds }
+          .forEach { it.toDate = LocalDate.now() }
```
iterative-it, line 48: Could be an opportunity now or in the future to deallocate the KW's allocations and clear their configuration as `staffConfigManager.removeStaffRole` currently does

src/test/kotlin/uk/gov/justice/digital/hmpps/keyworker/integration/AllocatableStaffSearchIntegrationTest.kt
cugorji-justice: yeah test were failing due to dublicated ids so wanted somthing unique but just for test

